PR #117, opened 2014-05-08
Use echo -e to enable enable interpretation of backslash escapes.

JuanitoFatas:
For https://github.com/bundler/bundler/issues/3018.


indirect:
Fixed this a different way via 42ee913 above.

PR #137, opened 2014-12-26
[WIP] Anchored titles in content.

simi:
Hello!

It will be nice to get content titles linkable via anchors. I prepared a small demo. Can anyone confirm this is good idea? I would like to finish it after.

screenshot-localhost 4567 2014-12-26 09-23-03

:christmas_tree:

### DIFF
--- a/helpers/anchored_text_helper.rb
+++ b/helpers/anchored_text_helper.rb
@@ -1,0 +1,8 @@
+require 'active_support/core_ext/string/inflections'
+
+module AnchoredTextHelper
+  def anchored_text(title)
+    octicon_link = '<span class="octicon octicon-link"></span>'
+    link_to(octicon_link, '', fragment: title.parameterize, id: title.parameterize, class: 'anchor') + title
+  end
+end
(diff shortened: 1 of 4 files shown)


simi:
Hello and happy new year!
FYI I'm finishing this. But I'm wondering if all haml files are maintained manually. Isn't there any process how to generate some of those files? For example `source/v1.7/bundle_binstubs.haml` looks like generated file.

Also I will probably need to remove current id parameter for various h2 tags ([example](https://github.com/bundler/bundler-site/blob/master/source/v1.7/bundler_sharing.haml#L1-L2)). I can't find reason for them anyway.

TimMoore:
AFAIK they are maintained by hand.

indirect:
@simi any progress on this? if possible, it would be great to keep the existing anchors, but I understand if that isn't possible. Thanks!

simi:
I'm thinking how to automate this to avoid manual maintenance.

moeffju:
@simi How about adding the anchors manually (or semi-automatically through editor features) for v1.7 and v1.8 while you look to fully automate this?

kruczjak:
It should be done in #218 :)

PR #142, opened 2015-03-07
use https://rubygems.org

chulkilee:
Since using HTTPS is much more secure than HTTP, it would be better to use HTTPS in the documentation.

I believe old versions of bundle also work with https source.


indirect:
Thanks!

PR #144, opened 2015-03-21
Update for v1.9

segiddins:
\c @indirect


Rakefile
```diff
@@ -27,7 +32,7 @@ task :man => [:update_vendor] do
   end
 
   # Make man pages for the latest version available at the top level, too.
-  cp_r "build/v1.8/man", "build/man"
+  cp_r "build/#{VERSIONS.last}/man", "build/man"
```
TimMoore: :heart:

source/v1.9/whats_new.haml
```diff
@@ -0,0 +1,65 @@
+= partial "shared/whats_new"
+
+%h2#version18 What's New in v1.8
```
TimMoore: Hey @segiddins it looks like the What's New page has been copied from 1.8 without any changes. You could probably mostly copy & paste it from the blog post.

segiddins: Gah I didn't know this page existed! Fixed :)

TimMoore:
Thanks! I'm giving it a quick look over now and then I'll merge & release to the site.

Rakefile
```diff
@@ -9,9 +9,14 @@ task :update_vendor => ["vendor/bundler"] do
...
 
+VERSIONS = %w(v0.9 v1.0 v1.1 v1.2 v1.3 v1.5 v1.6 v1.7 v1.8 v1.9).freeze
+task :versions do
+  puts VERSIONS.join(' ')
+end
+
 desc "Pull in the man pages for the specified gem versions."
 task :man => [:update_vendor] do
-  %w(v1.0 v1.1 v1.2 v1.3 v1.5 v1.6 v1.7 v1.8).each do |version|
+  VERSIONS.each do |version|
```
TimMoore: Drat! This actually breaks because v0.9 doesn't have man pages. It's included in the `VERSIONS` list because the other v0.9 docs are still linked from `source/partials/_footer.haml`.

@indirect what do you think about omitting v0.9 from the site from now on? Hopefully nobody still uses it, and I don't think semver would dictate that we have to keep supporting pre-releases.

Alternatively, we could hack in an exclusion for v0.9 from the man list, or an inclusion in the footer partial, but IMO we should just go with the simple solution of getting rid of it entirely.

In the interest of moving forward, I'll go ahead and pull it out of the `VERSIONS` array for now.

TimMoore: I won't actually remove the directory, so old links to the v0.9 docs should still work. There just won't be a link in the header.

TimMoore:
This is live on http://bundler.io/ now. Thanks, @segiddins! :smile:

PR #145, opened 2015-03-22
Move old docs links to their own page

EduardoBautista:


source/partials/_sidebar.haml
```diff
@@ -32,7 +32,8 @@
     %li= link_to 'Deploying', "/#{current_version}/deploying.html"
     %li= link_to 'Sharing', "/#{current_version}/bundler_sharing.html"
     %li= link_to 'Updating Gems', "/#{current_version}/updating_gems.html"
-    %li= link_to 'Compatible versions', '/compatibility.html'
+    %li= link_to 'Compatiblity', '/compatibility.html'
```
EduardoBautista: I changed this because I believed it looked awkward next to "Older Versions".  I can change this back it you think it's better the way it was.

EduardoBautista:
This fixes https://github.com/bundler/bundler-site/issues/138

indirect:
Thanks! After thinking about this a little bit more, can we keep the current version and the previous version at the top? I don't want an infinitely-expanding list, but it's definitely useful to make it easy to get to the docs for the most recent version before the current one.

EduardoBautista:
@indirect Sure.

EduardoBautista:
@indirect Should they be called "v1.8" and "v1.9" or "Current Version" and "Previous Version"?

TimMoore:
Why don't we have both?

indirect:
I guess it seems clear to me that 1.8 is the previous if it says v1.9 and v1.8 :)

EduardoBautista:
screenshot 2015-03-22 18 17 02

How's this?

indirect:
Fine with me.

EduardoBautista:
Done.

indirect:
thanks!

PR #149, opened 2015-03-30
typo(gemfile): from the same same -> from the same

threetreeslight:
It might be typo, is it correct?


skottler:
Thanks @threetreeslight!

PR #150, opened 2015-04-08
Use proper version number

marocchino:


indirect:
The text is actually correct as written—we stopped recommending multiple source lines "as of Bundler 1.7", meaning "since the time that Bundler 1.7 was released".

marocchino:
Okey. :3

PR #154, opened 2015-05-11
Fix content extending into the side bar

rgb-one:
On the bundle update command page, it can be observed that some of the content from the command block and some text below it leaks into the side bar causing the text to clash. This pull request aligns the text properly.


indirect:
Thanks!

PR #157, opened 2015-05-21
Rename 'Bundler Proposal and Rationale' to 'Using Bundler' and add so…

rgb-one:
…me content.

### DIFF
--- a/source/v1.9/index.haml
+++ b/source/v1.9/index.haml
@@ -1,17 +1,19 @@
 %h2 What is Bundler?
 
 %p#intro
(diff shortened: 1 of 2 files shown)


indirect:
Do you feel like the "rationale" page is a good explanation of how to use Bundler? I would expect to point people at the homepage and "learn more" pages instead... what do you think?

rgb-one:
The content could do with some updating but the information is quite useful.  "Learn More" seems like an adequate title to me.

indirect:
Sounds good to me.

PR #159, opened 2015-06-05
Explain that install --deployment is also suitable for CI

mattbrictson:
It wasn't clear to me that `--deployment` is also meant to be used for CI environments. This PR makes it clear and also explains a bit more what `--deployment` is doing.

### DIFF
--- a/source/v1.10/bundle_install.haml
+++ b/source/v1.10/bundle_install.haml
@@ -19,7 +19,7 @@
       %p
         <code>--clean</code>: Run bundle clean automatically after install
       %p
-        <code>--deployment</code>: Install using defaults tuned for deployment environments
+        <code>--deployment</code>: Install using defaults tuned for deployment and CI environments
       %p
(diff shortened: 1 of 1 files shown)


segiddins:
I'd simplify it, and say deployment should be used on _any_ server environment, but that's just my opinion

mattbrictson:
@segiddins CI is kind of a hybrid: yes, it is a server, but it also needs to bundle your development and test gems in order to e.g. run tests, do static analysis, etc. Hence the confusion (speaking for myself, here). That's why I think **CI** should be mentioned explicitly.

mattbrictson:
Related change to the man page: https://github.com/bundler/bundler/pull/3723

mattbrictson:
@segiddins @indirect Now that bundler/bundler#3723 (man page change) has been merged, any chance of this one getting merged as well?

I've rebased this PR such that the edits are applied to `source/v1.10` rather than `source/v1.9`.

indirect:
Yeah, this makes sense to me. Thanks!

PR #163, opened 2015-07-02
Fix confusing language in version 1.10 release post

parkr:
Upon reading the post, this piece confused me. `fixes definitely fix` read as 2 verbs instead of a noun, then verb. Using a different word helps clear this up to my mind.

Thanks!


indirect:
Thanks!

parkr:
Thanks, @indirect. Sorry that `BUNDLED WITH` has been such a headache and thanks for taking the time to explain your reasoning.

PR #166, opened 2015-07-18
Improve the documentation for `bundle gem` and surface it on the homepage

mattbrictson:
`bundle gem` is the de-facto standard for creating rubygem projects, but this is not readily apparent by reading the current bundler.io site. This PR adds a section to the homepage explaining the `bundle gem` command, and improves the existing documentation.

### DIFF
--- a/source/v1.10/index.haml
+++ b/source/v1.10/index.haml
@@ -85,7 +85,7 @@
 %h2#create-gem Create a rubygem with Bundler
 
 %p
-  Bundler is also the best way to create new gems. Just like you might create a standard Rails project using <code>rails new</code>, you can create a standard gem project with <code>bundle gem</code>.
+  Bundler is also an easy way to create new gems. Just like you might create a standard Rails project using <code>rails new</code>, you can create a standard gem project with <code>bundle gem</code>.
 
 .bullet
   .description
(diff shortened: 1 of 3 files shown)


source/v1.10/index.haml
```diff
@@ -82,6 +82,35 @@
       bundle, and will always work.
     = link_to 'Learn More: Executables', './man/bundle-exec.1.html'
 
+%h2#create-gem Create a rubygem with Bundler
+
+%p
+  Bundler is also the best way to create new gems. Just like you might create a standard Rails project using <code>rails new</code>, you can create a standard gem project with <code>bundle gem</code>.
```
segiddins: > Bundler is also the best way to create new gems.

That isn't exactly true, and sounds much more grandiose than the feature is :P

indirect: I'm definitely down with descriptions like "easiest" or "simplest", but I agree that "the best" is maybe too self-praising for our own homepage. :)

mattbrictson: How about?

> Bundler also comes with an easy way to create new gems.

mattbrictson:
OK, I've just added a commit to tone down the home page language from "*the best way*" to "*an easy way* to create new gems." If this looks good to merge, I can squash it down to a single commit.

indirect:
:+1: cool, thanks!

PR #167, opened 2015-07-18
Sidebar fix

rgb-one:
Fix sidebar not appearing on docs and blog pages.


indirect:
Thanks!

source/layouts/blog_layout.haml
```diff
@@ -1,5 +1,5 @@
 !!!
-- v = current_page.url.scan(/v\d\.\d/).first || current_version
+- v = current_page.url.scan(/v\d\.\d+/).first || current_version
```
segiddins: can we use the `Gem::Version::VERSION_PATTERN` constant here?

indirect: Oh hey, that's a good idea. :+1:

PR #168, opened 2015-07-19
Markdown extentions, set partials_dir, create base layout

rgb-one:
Highlights:
- Add markdown extention for features not available in standard markdown  
- Set partial directory for easy referencing. ie. `partial 'partials/header' to partial 'header'
- Create a base layout which all other layouts inherit from.


indirect:
A good start. :)

PR #170, opened 2015-07-24
Haml pre block indentation fix and add paragraph tags to blocks of text

rgb-one:
- Simple solution to haml adding indentation to `pre` blocks from markdown files. 

For an illustration of this, have a look at the code block on the "Report a bug" page on the bundler website. Every line after the first additional indentation. The solution above is mentioned here: http://stackoverflow.com/questions/16605840/multiline-code-block-in-markdown-adds-unwanted-tabs

- Add paragraph tags to block of text on index page

### DIFF
--- a/source/layouts/blog_layout.haml
+++ b/source/layouts/blog_layout.haml
@@ -15,5 +15,5 @@
             = current_article.date.strftime('%b %e %Y')
         %hr
         = find_and_preserve do
-          = yield
+          ~ yield
   = partial 'partials/footer', locals: { v: v }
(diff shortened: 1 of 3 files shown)


segiddins:
@rgb-one have a screenshot for comparison?

indirect:
:+1: the code looks good, a screenshot would make it easy to merge without having to run this PR on my local machine :)

rgb-one:
You can actually see it on  http://bundler.io clicking "Report a bug" on the sidebar.

indirect:
Cool! In that case, a screenshot with before/after showing the fix would be awesome.

rgb-one:
Before `~`, every line after the first has unspecified indentation
before

After `~`
after

indirect:
Perfect :+1: :ship:

PR #184, opened 2016-03-01
Fix typo of 'a' article before vowels

kolotaev:
I'm not a native English speaking person, but AFAIK "an" article is used in preference to "a" one when it goes before vowel letters or sounds. At least, my tongue stumbled when I passed that place while reading documentation. :)

Hope, this PR will be helpful.


segiddins:
Thanks!

PR #188, opened 2016-04-28
Fixes reference to wrong version on Rationale page

grzuy:


RochesterinNYC:
Thanks for the correction and PR!

PR #192, opened 2016-04-29
Correct `--tiny` flag to `--patch` flag for `bundle outdated`

RochesterinNYC:
- The flag is called `--patch` for targeting the third semver portion via `bundle outdated`: bundler/bundler#4312


segiddins:
👍🏻

PR #194, opened 2016-04-30
[Contributors] Move Smit and Terrence to alum

segiddins:
@Who828 @hone I think this is in line with what we discussed about a month ago?


hone:
Sure seems good. Thanks.

PR #195, opened 2016-05-01
Update middleman to v4.1

kruczjak:
It seems like middleman v4.1 needs files with .html (e.g. `index.html.haml`), so I had to change a lot of them.

and there is a problem with https://github.com/bundler/bundler-site/pull/195/files#diff-f0596fb5a5c16d29feb78ed015d78841R3
I'm not sure, why it's failing build.


source/blog/calendar.html.haml
```diff
@@ -0,0 +1,16 @@
+%h2
+  Bundler news and updates from
+  - if defined?(page_type) #TODO ? fails when building without this line
```
indirect: Sounds like they removed the `page_type` method in Middleman 4. According to the [middleman-blog docs](https://middlemanapp.com/basics/blogging/) you can run `middleman init` to re-generate the calendar page. Maybe that will show you what to do instead?

indirect:
Renaming the files seems fine, so once you have the calendar page upgraded, rebase this on master, and I think we're good!

source/blog/calendar.html.haml
kruczjak: I was missing `blog.calendar_template = 'blog/calendar.html'` (it clearly started work after changing `calendar.haml` -> `calendar.html`)

PR #198, opened 2016-05-07
Warn about two factor auth and tokens for password input in site deploy rake task

RochesterinNYC:
- This caveat is not clear from the script and failure to enter in the correct credentials at this point in the deploy script can result in multiple commits being pushed in the eventual deploy.


segiddins:
👍

Issue #199, opened 2016-05-09
Redesign


LynnCo:
![bundler_blog](https://cloud.githubusercontent.com/assets/5768468/15129449/80105db4-15f6-11e6-8bcb-1317b0e770c8.png)
![bundler_homepage](https://cloud.githubusercontent.com/assets/5768468/15129450/8021f146-15f6-11e6-8cdc-0d992ec29c2e.png)
![bundler-site-header](https://cloud.githubusercontent.com/assets/5768468/15129451/802dc05c-15f6-11e6-82ee-6b2b3d9d5da9.png)

kruczjak:
Closed by #218

Issue #202, opened 2016-05-09
Sidebar for the documentation view with command reference and all other documentation content


LynnCo:
Related https://github.com/bundler/bundler-site/issues/199

kruczjak:
Closed by #218

Issue #201, opened 2016-05-09
Top nav with most important navigation links for the site


LynnCo:
related https://github.com/bundler/bundler-site/issues/199

kruczjak:
Closed by #218

PR #218, opened 2016-05-25
Redesigning website

kruczjak:
Redesigning + plugin to minify html + started i18n localization.


kruczjak:
This PR should close #199, #201 and #202

kruczjak:
It's only missing "Docs" header on some pages :)

If you see anything wrong/anything what could be done better, let me know :)

segiddins:
On mobile, the buttons should probably be centered instead of right-aligned

source/contributors.html.haml
```diff
@@ -1,73 +1,46 @@
...
+          %h4
+            %b James Wen
+          %h5= link_to '@RochesterinNYC', 'https://github.com/RochesterinNYC'
+
+    .col-xs-6.col-md-4
+      .thumbnail
+        = image_tag 'team/sam_hexagon.png'
+        .caption.text-center
+          %h4
+            %b Sam Geddins
```
segiddins: Samuel Giddins

source/contributors.html.haml
```diff
@@ -1,73 +1,46 @@
...
+      .thumbnail
+        = image_tag 'team/james_hexagon.png'
+        .caption.text-center
+          %h4
+            %b James Wen
+          %h5= link_to '@RochesterinNYC', 'https://github.com/RochesterinNYC'
+
+    .col-xs-6.col-md-4
+      .thumbnail
+        = image_tag 'team/sam_hexagon.png'
```
segiddins: samuel

segiddins:
Could you add a script to hexagon-ify a team member image?

sailorhg:
hi! this is looking great, thank you for building it! here are some design notes:

- [ ] i'll design a new footer to put in

- [ ] it looks like the titles are all an equal width but that means that the text is a different size in each one. i'd calculate the pixel widths as a % of the original and have the % consistent across title images, not the pixel widths. and i think they'd all look better a little smaller than they are right now. for example i think the docs header looks good as 400px wide.

- [ ] the blog should have the first entry before the list of previous posts

- [ ] the docs page should have more vertical padding before the text, (and more space before the footer) and the columns should be centered which they look like they are at some page widths but not others. example:
screen shot 2016-06-09 at 3 26 18 pm

- [ ] i'd include the docs header in each of the detailed docs pages

- [ ] for the team page i'd resize the hexagons to be a little smaller (this is likely my bad if that was the size they were in the design). maybe 200px? i'd also put more padding between the header and the content

- [ ] right now the header with the titles changes height depending on the title image height. the title images all vary based on the size of the icon. i'd keep the header a fixed height.

i'm super excited to see this go live! thank youuu!

kruczjak:
@sailorhg Hi :)
I've just added commit with proposed changes from you :)

If you see something is still wrong or can be better, let me know :)

sailorhg:
@kruczjak thanks for your changes! here's another iteration of feedback:

- [ ] footer: the new footer should be in the bundler_homepage.png file in the folder with all the other designs

bundler_homepage

- [ ] docs page: thanks for centering the 3 columns on the docs page! the text of the columns themselves can be left-aligned though

- [ ] blog page: i've updated the design folder with a design for including the full text of the most recent blog post:

- [ ] page headers: sorry if i was unclear about this in my previous feedback. even if all the titles are the same width, their text is currently different sizes because the titles are of varying lengths. would it be possible to keep the scaling of the titles consistent, instead of the pixel width? the docs title image is the correct width. but the other titles shouldn't necessarily be 400px, they can be the width that means they and the docs title image have been scaled the same amount and the text, if super imposed would be the same height.

kruczjak:
New changes should be deployed :)

About page headers, I couldn't make them as % of height/width because font in these images haven't got same pixel size on everyone, but I've tried to set the best settings for each header independently :)

And I'm not quite sure about columns in "Docs" page, so I'm waiting for feedback :)

indirect:
@kruczjak @sailorhg this is looking better all the time! :tada: I noticed a couple of things while checking out the current preview:

- [ ] at the largest responsive size, the font size is too small. can you increase it by 10-20%?
- [ ] the text inside the blue areas (like "Getting started with Bundler is easy!") is even smaller than the paragraph font size. can you enlarge it to match the paragraph font size?
- [ ] on pages other than the home page (like `/v1.12/index.html`), the buttons are not "button-y". The text like "Learn More: Gemfiles" is just a plain link, and not a blue button like it is on the homepage.
- [ ] on the Docs index page, the sub-headers (like "Utilities") should be left-aligned, and the versions (like "v1.12") should be left aligned, and would probably look better fully written out (like "Version 1.12")

Thanks! :D

sailorhg:
@kruczjak @indirect yes it keeps looking better and better every time!

besides @indirect's feedback, the last piece of feedback i can think of:

- [ ]  can the width of the blog text for the most recent blog entry take up like 70 or 80% of the page instead of the whole width of the page? it doesn't have to be exactly 70 or 80% it can be the closest grid column that is about that width.

thanks!

indirect:
Oh, I remembered one more thing: when <code> is inside a header (like <h1>) I think it would be better to remove the red background and red color. Let's keep it black text on transparent background, and let the font change show what is code inside headers.

indirect:
Almost!

kruczjak:
Ups 😭 
Should be ok now :D

sailorhg:
sorry, last-last feedback! do you mind changing the headers on the docs page to be left aligned, and then changing the distribution of the columns to be 5, 4, 2? i think this might look more balanced since the last column has such short words in it.

when i changed it in the inspector it ended up looking something like this: 

screen shot 2016-06-16 at 2 51 55 pm

indirect:
👍

kruczjak:
Done :)

indirect:
On single column view, it goes back to center-aligned, but I think it should be left-aligned:

sailorhg:
i just noticed that this happens to the team page on mobile widths: 
screen shot 2016-06-21 at 1 34 07 pm

but i checked all other pages and they look great!

andremedeiros:
@indirect on a side note, we look great! ❤️

kruczjak:
@sailorhg, it should be ok now :)

sailorhg:
@kruczjak everything looks great to me!! thanks so much for your work on this!

segiddins:
@kruczjak the team images still appear to be cropped hexagons -- I thought we were moving to doing that with CSS?

indirect:
I think we were going to use an svg outline to crop circular photos in the future? (But the current photos are already cropped to be hexagons).

kruczjak:
Yes and there are CSS and example for doing this :)

sailorhg:
the dropbox assets folder contains all the team photos as squares for whenever you want to make that change!

kruczjak:
So, I've just changed hexagon images to rectangular and cropped with svg :)
I've left both versions to compare if it works ok :) (first avatar is cropped file and second next to it is cropped with svg).

I'm waiting for feedback and then I will remove old ones. :)

source/stylesheets/_hexagon_image.scss
```diff
@@ -20,12 +20,8 @@
...
   }
-}
-
-.polygon-clip-hexagon {
-  -webkit-clip-path: polygon(50% 0%, 100% 25%, 100% 75%, 50% 100%, 0% 75%, 0% 25%);
-  clip-path: polygon(50% 0%, 100% 25%, 100% 75%, 50% 100%, 0% 75%, 0% 25%);
-  -webkit-clip-path: url("#polygon-clip-hexagon");
-  clip-path: url("#polygon-clip-hexagon");
-}
+}
```
segiddins: missing trailing newline

indirect:
It looks great to me! 👍

kruczjak:
Great :D I will remove duplicated images :)

zzak:
@sailorhg @kruczjak @indirect Great redesign team, looks great! But where did the alum go? :( http://bundler.io/contributors.html

I am pretty sure there would be no bundler w/o @wycats and @carllerche :bow:

indirect:
@zzak the current page just has the current team, but we plans to add credits for all current and past contributors! Stay tuned. :)

wycats:
@zzak Another major alum: @hone. People probably won't forget that I wrote the original versions, but they might well forget how instrumental @hone was.

indirect:
You may be interested in #225.

PR #223, opened 2016-07-14
Fix contributors page rendering for Safari

maxjacobson:
I noticed that the contributors page wasn't rendering correctly in
Safari on my iPhone -- the first team member's face/name show ups, and
then after that there is only white, no other team members or footer.

I think the issue is that there are multiple, conflicting strategies for
"clipping" the profile photos into hexagons, and there's some kind of
clash there.

With this change, the page renders correctly for Safari and Chrome. On
Firefox, the page will render, but it no longer has hexagon-clipped
profile photos. According to http://bennettfeely.com/clippy/ the
clip-path property doesn't work in Firefox, although the previous
implementation did manage to get it working by embedding an svg in the
HTML and then loading that.

IMO losing the hexagon-clipping on Firefox is worth it for the page to
render in Safari, although admittedly that calculus is subjective.

(I also haven't checked to see how other browsers are affected.)

### DIFF
--- a/source/contributors.html.haml
+++ b/source/contributors.html.haml
@@ -4,14 +4,6 @@
     class: 'img-responsive header-padding',
     style: 'max-width: 460px; padding-bottom: 10px; width: 78%;'
 
-:css
-  .polygon-clip-hexagon {
-    -webkit-clip-path: polygon(25% 6%, 75% 6%, 100% 50%, 75% 94%, 25% 94%, 0% 50%);
-    clip-path: polygon(25% 6%, 75% 6%, 100% 50%, 75% 94%, 25% 94%, 0% 50%);
-    -webkit-clip-path: url(#polygon);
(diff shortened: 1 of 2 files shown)


indirect:
Whoa, that's really weird! The teams page worked perfectly in Safari for months... including (I think) briefly after we first deployed it. Let me see if we have a previous revision that worked.

indirect:
Hmm... @maxjacobson, looks like you're right, and it was be57b6e9ee931c71caff23d04bbc07b4f7bd0c45 that broke the teams page in Safari. Thanks for tracking this down and submitting a fix!

@kruczjak, do you think you could do another PR that adds back the SVG cropping, but only for Firefox?

maxjacobson:
Cool, glad to help :)

kruczjak:
@indirect yes :)

PR #225, opened 2016-07-15
Add all contributors to team page

indirect:
The list of contributors is pulled from the github API during the full build process.


zzak:
Hello @indirect and thanks for taking the time to work on this.

While it's good to show appreciation to all contributors, I don't feel this properly pays respects due to the core alumni by combining their efforts with regular contributors based on number of commits.

In addition to this, I think there should still be a "former core alumni" section of the team page and would recommend using the [Rails community page](http://rubyonrails.org/community/) as inspiration.

Thank you!

indirect:
@zzak thanks for the feedback!

segiddins:
👍

Issue #207, opened 2016-05-09
execute the documentation code examples during the test suite


lynncyrin:
the documentation (both the manpages/help and the website at bundler.io) contains many examples of running commands. we should run those examples as part of the test suite and make sure they work!

esasse:
After some [discussion on Slack](https://bundler.slack.com/archives/C0ZRUQ78V/p1576326786001000) we decided to close old issues with ideas for improvement.

Please, feel free to reopen it if you intend to work on it in the foreseeable future.

PR #242, opened 2016-08-14
Tests generator from documentation

kruczjak:
Tests generator for #207 

I'm still not sure what to do with those generated specs :)

Example generated spec:

```
require 'rspec'

describe 'build/man/bundle-install.1.html' do
it '#0' do
  bundle! 'install'
end
end
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ source/conduct.html.md
 # man pages built from the bundler repo
 /source/man/
 /source/v*/man/
+
+# rspecs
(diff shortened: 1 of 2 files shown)


segiddins:
Looks like we'll need to pull in the gemfile that's being installed as well

segiddins:
`171 examples, 82 failures`

indirect:
Closing this for now, still maybe worth pursuing more later when someone has time. 👍

PR #259, opened 2016-09-20
Fix broken links to "Learn More: Packing"

dentarg:
Aims to fix the broken `Learn More: Packing`links on http://bundler.io/deploying.html and https://bundler.io/v1.12/deploying.html


lynnco:
Thank you!

PR #279, opened 2016-11-22
FAQ on ruby RUBY_VERSION trick

olleolleolle:
This PR tries to explain the use of the `ruby RUBY_VERSION` "trick" to the gem author audience.

This is a companion PR to https://github.com/bundler/bundler/pull/5200


indirect:
I think (hope?) that this is now fixed without the less-than-ideal workaround. Thank you for documenting this, though!

PR #295, opened 2017-03-10
Fix Getting Started link to bundle_gem man pages

### LINKED ISSUE #294: http://bundler.io/bundle_gem.html gets sent to a "Page not found"
Hi!

I really like bundler, and the site http://bundler.io/ is really helpful! thanks for that!

the issue is when you start from http://bundler.io/v1.12/#getting-started

then click the button about making your own gem `Learn More: bundle gem`

you are take to http://bundler.io/bundle_gem.html  which has no content on it :(

it looks like there should be content for this page ( based on the code )
https://github.com/bundler/bundler-site/blob/master/source/v1.11/bundle_gem.html.haml

theoretick:
Fixes #294

### DIFF
--- a/source/v1.12/index.html.md
+++ b/source/v1.12/index.html.md
@@ -98,7 +98,7 @@ Create a new gem with a README, .gemspec, Rakefile, directory structure, and all
           create  my_gem/test/my_gem_test.rb
     Initializing git repo in ./my_gem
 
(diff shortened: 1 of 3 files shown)


indirect:
awesome, thanks so much!

indirect:
@bundlerbot r+

PR #299, opened 2017-04-10
Add troubleshooting page to vendored site from bundler

jules2689:
Problem:

https://twitter.com/rappmasterflash/status/851494907906670592

~The troubleshooting page contains no other links to other pages, though I suspect that this could happen in the future. Perhaps things need to be a bit more intelligent? Could the vendoring convert all relative links to absolute ones instead?~

This copies over the entire `doc` directory to be under `bundler.io/doc/*`. This means that `issues` is no longer treated as a special snowflake, although Code of Conduct still is.

A file at `doc/path/NAME_OF_FILE.md` would be at `bundler.io/doc/path/name_of_file.html` in the end.


source/issues.html.md
```diff
@@ -2,7 +2,7 @@
 
 So! You're having problems with Bundler. This file is here to help. If you're running into an error, try reading the rest of this file for help. If you can't figure out how to solve your problem, there are also instructions on how to report a bug.
 
-Before filing an issue, check our [troubleshooting guide](../TROUBLESHOOTING.md) for quick fixes to common issues.
+Before filing an issue, check our [troubleshooting guide](https://github.com/bundler/bundler/blob/a08efb41bc82c6df91834e8dbd06574f4368dbb4/doc/TROUBLESHOOTING.md) for quick fixes to common issues.
```
andremedeiros: Shouldn't this always point to `master` instead? This so people always get the updated version.

lib/tasks/vendor_files.rake
```diff
@@ -9,11 +11,30 @@ end
...
-    cp "doc/contributing/ISSUES.md", "../../source/issues.html.md"
-    cp "CODE_OF_CONDUCT.md", "../../source/conduct.html.md"
+    sha = `git rev-parse HEAD`.strip
+
+    # For each path read the file, gsub the content for relative links, output to "to" path
+    paths.each do |from, to|
+      content = File.read(from)
+      content.gsub!(relative_link_regex) do |match_data|
+        # Both paths are relative, so calculate the actual path using Ruby's pathname
+        from_path = Pathname.new(File.dirname(from))
```
segiddins: `Pathname.new().dirname`

segiddins:
Won't this mean we're linking to a page that no longer has the bundler.io styling?

jules2689:
@segiddins it means we're linking to Github instead. I figured this may be a good solution as we don't know how deep the linking will go (i.e. Issues references troubleshooting right now, but someone may add a link to another page in the future... do we just keep adding anything?)

source/issues.html.md
jules2689: I had chosen to make the vendor rake task use a sha so that a website is linked to a specific public version. Scenario:
1. bundler site is published on bundler sha 12345
2. bundler/bundler is updated, the page changes. Require full site deployment to update docs.
3. In the meantime, we are pointing to some new docs and some old docs

colby-swandale: We want to point to master so that we're giving up to date documentation to the user.

segiddins:
I don't think we want to point to master, this way we can update docs for new (unreleased) versions

jules2689:
yea, what @segiddins said was my thought process. If we have differing docs on released version and master, we'd have a conflict.

jules2689:
ping @segiddins @colby-swandale thoughts on proceeding with this?

segiddins:
I kinda like having the markdown files on the site, they look nicer than on github (i.e. http://bundler.io/conduct)

jules2689:
If we'd like to do that, what are your thoughts on just publishing the entire subdirectory of `docs` on the website? Otherwise some page somewhere will end up relatively referencing a page that isn't published outside of github

segiddins:
If @feministy is good with that, I think it'd be excellent

indirect:
yeah, let's put the entire docs directory on the site, and rewrite relative links so that the site links to itself 👍

jules2689:
This has been updated to copy over the entire `doc` folder. I'm a bit confused about the gitignore, which had `source/issues` and `source/conduct` gitignored, but they were committed anyways. Anyone wanna shine some light on that?

lib/tasks/vendor_files.rake
```diff
@@ -12,8 +49,15 @@ task :repo_pages => [:update_vendor] do
   Dir.chdir "vendor/bundler" do
     sh "git reset --hard HEAD"
     sh "git checkout origin/master"
-    cp "doc/contributing/ISSUES.md", "../../source/issues.html.md"
-    cp "CODE_OF_CONDUCT.md", "../../source/conduct.html.md"
+    sha = `git rev-parse HEAD`.strip
```
segiddins: This is now unused?

jules2689: Good catch 👍

segiddins:
If you rebase this, we should be able to get a heroku review app that will let us see the generated changes

jules2689:
Rebased, let's give it a shot :+1:

lib/tasks/vendor_files.rake
```diff
@@ -1,3 +1,40 @@
...
+
+# Match
+# ([get an invite here](contributing/GETTING_HELP.md))
+# [get an invite here](contributing/GETTING_HELP.md)
+# Dont Match
+# [Heroku support](https://www.heroku.com/support)
+# [Heroku support](http://www.heroku.com/support)
+RELATIVE_LINK_REGEX = %r{
+  \[(?<title>.*)\]            # Match title
+  \((?<link>[^(http)].*?)\)   # Match all non-http links
```
segiddins: I don't think `[^(http)]` is what you want here

segiddins:
Lots of build errors https://dashboard.heroku.com/apps/bundler-site-staging-pr-299/activity/builds/4321e9f6-e5b5-4a5c-b1f6-3e539859a700

lib/tasks/vendor_files.rake
jules2689: This will capture it, but we ignore it. I could use `[^(?:http)]` which is basically saying "non-http*" links.

That work?

jules2689:
Closing in favour of https://github.com/bundler/bundler-site/pull/317

PR #300, opened 2017-04-11
Use up-to-date, smaller logos

j-delaney:
Right now the Sponsors page is using rather large logos for both Pagerduty and Stripe so that it looks like this:

Huge Stripe logo

Huge PagerDuty logo

This PR uses smaller logos (500px wide) that are also up-to-date (both Pagerduty and Stripe have updated their logos since they were first added here).

Disclaimer: I work for Stripe.


j-delaney:
This is what the Sponsors page looks like with my changes:

Whole Site

colby-swandale:
Thank you ❤️ (Sorry this took so long)

colby-swandale:
@bundlerbot r+

PR #310, opened 2017-06-20
[wip] RubyGems.org SSL troubleshooting guide

indirect:
This guide is a work in progress. Please don't merge it yet.

Today, users of Bundler and RubyGems can run into SSL errors if their system CA cert bundle is old, or if they are running an older version of Bundler/RubyGems. By January 2018, users will start to see errors if their Ruby is linked against Openssl < 1.0.1, because that is when we will start to require TLSv1.2.

After consultation with @rubymorillo, our diagnosis was that the existing RubyGems SSL guide is both a confusing patchwork of different writeups and also out of date.

Our fix is to write comprehensive Bundler/RubyGems guide(s), explaining why we use SSL in the first place, what the underlying cause of the various SSL errors is, and suggestions for how to possibly fix those errors, as well as links to other potential resources.

Writing better docs and being able to refer to them sounds wayyyy better than trying to help out users with the same problems one at a time.

### DIFF
--- a/source/localizable/v1.15/guides/how_bundler_uses_ssl.en.html.md
+++ b/source/localizable/v1.15/guides/how_bundler_uses_ssl.en.html.md
@@ -1,0 +1,24 @@
+# How Bundler uses SSL certificates
+
+## Introduction
+
+All web browsers come with a certificate authority (CA) bundle. These are cryptographic certificates provided by the companies that sell SSL certificates (like Verisign, Globalsign, and others). Using those "root” keys/certificates (they’re called root keys because they are the keys from which many other companies and websites derive their SSL certificates), web browsers “know” they can trust the SSL certificate being given to them by a particular website, such as RubyGems.org.
+ 
+Occasionally, new companies are added to the CA bundle, or existing companies have their certificates expire and need to distribute new ones. For most websites, this isn't a huge problem, because web browsers regularly update their CA bundle as part of general browser updates. 
+ 
+## How Ruby uses CA bundles
+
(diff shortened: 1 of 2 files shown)


rubymorillo:
Hey team, any idea when this'll get merged? (I could, but I don't want to since code reviews aren't my thing). I could use the live link to get started on the PR for the updated RubyGems SSL guide. Thank you 😄 
cc @segiddins @kruczjak

segiddins:
@indirect please add a PR description explaining what the guide is for (and also rebase, since two of the commits are already on master)

indirect:
@rubymorillo This SSL guide has a link to the troubleshooting guide, but it doesn't go anywhere yet because we didn't write the troubleshooting guide yet. 

I was initially thinking that we would add the troubleshooting guide to this PR at the same time, but after our chat today I'm thinking that maybe we should expand this page into a longer guide, containing 1) an explanation of why and how we use SSL, 2) diagnosis steps to figure out if an error you are seeing is because of old certs, or because of TLS versions, or because of something else, and 3) suggestions to resolve these errors, including how to replace certs, how to manually upgrade rubygems, and how to upgrade openssl+ruby to a new version. (Oh and for users who are having some other kind of SSL problem, we should probably refer them to Mislav's Ruby and SSL troubleshooting guide.)

Does that seem reasonable?

rubymorillo:
@indirect yep, that sounds reasonable! I guess we can close this out and I'll get started on expanding it :)

indirect:
Closing this as superseded by #321

PR #311, opened 2017-06-20
update middleman dependencies

indirect:
Updating rb-notify seems to resolve an issue where middleman would hang and stop responding to ctrl-C.


indirect:
@bundlerbot r+

PR #312, opened 2017-06-23
remove livereload

indirect:
it is apparently unmaintained, and because it triggers a full reload for
every single html file anytime any html file is changed, our site causes
it to get stuck livereloading every single file D:


indirect:
@bundlerbot r+

segiddins:
Gotta commit the lockfile change too

indirect:
thank you for saving me, @bundlerbot. r+

thanks for noticing, @segiddins 😅

PR #319, opened 2017-07-29
Add Spanish locale

arbonap:

The problem was...
- There's no localization of the bundler-site in castillian Spanish.

My diagnosis was...
- I saw that there was no `es.yml` file in the locales directory of the `bundler-site` repo.

My fix...
- I wanted to create the first localized Spanish file! ✨

I chose this fix because...
- I know we don't want to start localization of documentation until we add more explanations to the English dos, but this file seemed like a great first file to localize. It doesn't seem like it will change very much in spite of widespread doc changes.
💙

### DIFF
--- a/locales/es.yml
+++ b/locales/es.yml
@@ -1,0 +1,27 @@
+---
+es:
+  home:
+    bundler_info1: |
+      Bundler provee un ambiente consistente para los proyectos de Ruby por modo de siguiendo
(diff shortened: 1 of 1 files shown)


indirect:
Awesome, thank you! <3

segiddins:
@indirect please use bundlerbot to merge

PR #320, opened 2017-07-31
Add git bisect explanation to "How to git bisect in projects using Bundler" guide

arbonap:

The problem was...
- I saw there was no background explanation of the `git bisect`command in the "How to `git bisect` in projects using Bundler" docs guide.

My diagnosis was...
- I added a new section at the beginning of the "How to `git bisect` in projects using Bundler" guide background information about what `git bisect` does.

My fix...
- I wrote up an introductory explanation of `git bisect`, and linked to the official git documentation, as well as to a Thoughtbot blog post that guides the user on how exactly to use `git bisect` with examples.

I chose this fix because...
- I thought this guide in particular would be less intimidating to anyone who is not familiar with the `git bisect` command if they are given some background explanation. 🍒

- Please let me know if you think this background explanation shouldn't go in its own section, and if it should rather be interspersed throughout the guide.
Happy to incorporate any changes!

### DIFF
--- a/source/v1.14/git_bisect.html.md
+++ b/source/v1.14/git_bisect.html.md
@@ -2,6 +2,12 @@
 title: How to git bisect in projects using Bundler
 ---
 
+## How to use git bisect
+
+[`git bisect`](https://git-scm.com/docs/git-bisect) is a useful debugging tool. For context, `git bisect` is a git command that can be used to track down the specific commit which a bug was introduced into the codebase.
+
+If you can find a commit where the code works properly and a commit with the offending bug, you don’t have to trace down the buggy commit by hand. The `git bisect` command, via binary search, will help you find the offending commit. For example, the Git documentation has [a handy `git bisect` guide](https://git-scm.com/book/en/v2/Git-Tools-Debugging-with-Git) that shows two ways to use it.
+
 ## How to git bisect in projects using Bundler
 
 A few things that may not be obvious are needed for `git bisect` to work


source/v1.14/git_bisect.html.md
```diff
@@ -2,9 +2,13 @@
 title: How to git bisect in projects using Bundler
 ---
 
+## How to use git bisect
+
+[`git bisect`](https://git-scm.com/book/en/v2/Git-Tools-Debugging-with-Git) is a useful debugging tool. For context, `git bisect` is a git command that can be used when a bug has been introduced into a codebase and you need to track down the specific commit that bug was introduced in. If you can find a commit where the code works properly and a commit with the offending bug, you don’t have to trace down the buggy commit by hand. The `git bisect` command, via binary search, will help you find the offending commit. [This](https://robots.thoughtbot.com/git-bisect) is a handy guide that shows two ways how to take advantage of `git bisect`.
```
colby-swandale:
```
[`git bisect`](https://git-scm.com/book/en/v2/Git-Tools-Debugging-with-Git) 
```

I would link this to https://git-scm.com/docs/git-bisect instead and move the debugging guide to where you linked the thoughtbot article.

colby-swandale:
```
[`git bisect`](https://git-scm.com/book/en/v2/Git-Tools-Debugging-with-Git) is a useful debugging tool. For context, `git bisect` is a git command that can be used when
```
Lots of starting and stopping here, can i suggest just making this: `git bisect is a git command that can be used...`

colby-swandale:
```
git command that can be used when a bug has been introduced into a codebase and you need to track down the specific commit that bug was introduced in
```
I would rearrange this slightly as:
```
git command that can be used to track down the specific commit which a bug was introduced into the codebase
```

colby-swandale: ` that bug was introduced in. If you can find a commit where`

I would add a paragraph here to help with readability

arbonap: Sounds good to me, thanks for the review @colby-swandale 👌🏼

source/v1.14/git_bisect.html.md
```diff
@@ -2,6 +2,12 @@
 title: How to git bisect in projects using Bundler
 ---
 
+## How to use git bisect
+
+[`git bisect`](https://git-scm.com/docs/git-bisect) is a useful debugging tool. For context, `git bisect` is a git command that can be used to track down the specific commit which a bug was introduced into the codebase.
+
+If you can find a commit where the code works properly and a commit with the offending bug, you don’t have to trace down the buggy commit by hand. The `git bisect` command, via binary search, will help you find the offending commit. [This](https://git-scm.com/book/en/v2/Git-Tools-Debugging-with-Git) is a handy guide that shows two ways how to take advantage of `git bisect`.
```
indirect: Instead of linking to "this", can we restructure so the link is a meaningful description? What do you think of "For example, the Git documentation has [a handy `git bisect` guide](...) that shows two ways to use it."

segiddins:
@indirect please use bundlerbot to merge

PR #321, opened 2017-08-04
add rubygems ssl troubleshooting guide

colby-swandale:
This PR adds a RubyGems.org TLS/SSL troubleshooting guide which explains the problems users may encounter relating to TLS/SSL and provides steps on solving there problem.

\cc @indirect


indirect:
Merged in #322.

PR #323, opened 2017-08-07
Fixing guides helper and guides url helper

colby-swandale:

Currently the guides are not being linked correctly in the sidebar and the `guides` helper is breaking in development due calling a method on `nil`

Clicking any link to the a guide in the sidebar does not go the intended page. And going to `docs.html` would break in development.

Remove `nil`s in the guides helper and remove the extra `/` in the guides url helper.


indirect:
@bundlerbot r+

Issue #326, opened 2017-08-11
Delete duplicate title on "What's New in each Release page"


arbonap:
I noticed on the [bundler.io documentation site](http://bundler.io), there needs to be a copy change.

It would be rad to delete the duplicate "title: What's new in v1.12" on the [What's New in each Release page for v1.12](http://bundler.io/v1.12/whats_new.html). Looks like this copy change should be made in the `bundler-site/source/v1.12/whats_new.html.haml` file

![screen shot 2017-08-10 at 7 50 15 pm](https://user-images.githubusercontent.com/15078895/29200415-f0d86c4a-7e09-11e7-82ee-999ae65b462f.png)

arbonap:
Cool, gonna close this since the change has been merged in ✨

PR #329, opened 2017-08-27
Remove duplicate header in What's New post

arbonap:

The problem was...
- There was a duplicate header for Bundler's "What's New in Each Release" v1.12 update. 

My diagnosis was...
- To get rid of it

My fix...
- Delete the unnecessary header.

I chose this fix because...
- Resolves Issue #326


segiddins:
@bundlerbot r+

PR #330, opened 2017-08-28
Fix broken link in 'What's New in v1.12'

arbonap:

The problem was...
- While reading through the "What's New in V1.12" of Bundler blog post, I noticed there was a broken link and some missing markup. 
It looked like this:
screen shot 2017-08-27 at 7 39 57 pm

My diagnosis was...
- I guess markdown syntax doesn't work in the `haml` blog post files. 

My fix...
- I added in vanilla HTML into the `haml` blog post file to add in the markup and fix the broken link.
- It should look like this, now:
screen shot 2017-08-27 at 7 39 09 pm

I chose this fix because...
- to improve the user's experience on the bundler.io site ✨

### DIFF
--- a/source/v1.12/whats_new.html.haml
+++ b/source/v1.12/whats_new.html.haml
@@ -15,15 +15,14 @@ title: What's New in v1.12
     .bullet
       .description
         %p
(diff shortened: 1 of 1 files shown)


segiddins:
@bundlerbot r+

PR #333, opened 2017-09-15
Fixed typo in PR template.

hmistry:

There is a typo in the next question...

Wrong word

Fixed with correct word

I couldn't think of another fix.


colby-swandale:
thanks! @bundlerbot r+

PR #336, opened 2017-09-15
Tweaked CSS to increase spacing a tad between paragraphs so there's a…

hmistry:
… bit more visual separation. This makes it easier to identify when reading.

I found the visual spacing between paragraphs to be too close to the line spacing making it harder to distinguish the different paragraphs.

Paragraph margin was too small.

Increase paragraph margin.

It's the simplest fix.

### DIFF
--- a/source/stylesheets/application.css.scss
+++ b/source/stylesheets/application.css.scss
@@ -134,6 +134,10 @@ h1, h2, h3, h4, h5 {
   }
 }
 
+p {
+  margin-bottom: 15px;
+}
+
 .contents {
   font-size: 17px;
 


colby-swandale:
Thanks for opening a pull request and helping improve bundler.io. Is there an specific example that motivates this change?

I'm asking because the difference in the bottom margin for `p` tags in this PR is only 4px with what is currently in master

hmistry:
@colby-swandale Master is currently at 11px which is same as line-spacing. Ideally you want it to be a bit more than line spacing, and even more than what I used but I'm limited by other header margins which is around 21px (?). So 15px was a quick in between value without having to change all the other header spacings.

Compare the attached images - see how Bundler's article the paragraphs kind of blend together where as the other article there's sufficient white space between paragraphs and even larger white space between paragraphs and the next heading. The latter gives us visual cues as to the grouping hierarchy of the text blocks.


**Bundler current spacing**
screen shot 2017-09-15 at 9 48 50 am


**Article with more white space between paragraphs**
screen shot 2017-09-15 at 9 51 41 am

hmistry:
Where do you suggest I change it? If in Bootstrap's CSS file, then what's to prevent it from getting overwritten when we upgrade Bootstrap?

esasse:
ping @colby-swandale do we still want to merge this?

duckinator:
It looks like the place the current `margin` for the `p` element is defined is in bootstrap itself, so this seems to be the most appropriate place for it.

I played with it locally real quick and agree that it makes it easier to read.

PR #337, opened 2017-09-18
Add quality spec for bundler.io docs

arbonap:

The problem was...
- Except for the man page documentation, there are no specs that test the quality of documentation on the [bundler.io](https://bundler.io) site. 

My diagnosis was...
- I added a quality spec that includes checking the quality of the blog posts and blog updates. Most notably, it sifts out gendered language as well as _actually_'s and other similar language so that we don't accidentally disrespect our reader. 

My fix...
- Create a script spec that sifts out unnecessary and non-inclusive language in the bundler documentation site's blogposts and guides. 

I chose this fix because...
- I spent a while trying to research ways to setup and incorporate RSpec into the [bundler.io](https://bundler.io) site but documentation related to trying to set-up RSpec with Middleman was kind of lacking. Plus, it doesn't seem like many people use RSpec with Middleman to begin with. This is a light-weight alternative.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -21,6 +21,7 @@ GEM
       coffee-script-source
       execjs
     coffee-script-source (1.12.2)
+    colorize (0.8.1)
     compass (1.0.3)
       chunky_png (~> 1.2)
       compass-core (~> 1.0.2)
@@ -207,6 +208,7 @@ PLATFORMS
 DEPENDENCIES
   bootstrap-sass
   builder
+  colorize
   haml (~> 4.0.7)
   jquery-middleman
   kramdown
@@ -230,4 +232,4 @@ RUBY VERSION
    ruby 2.4.1p111
 
 BUNDLED WITH
-   1.15.1
+   1.15.4
--- a/spec/quality_spec.rb
+++ b/spec/quality_spec.rb
@@ -1,0 +1,42 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'colorize'
+require 'pathname'
+require 'time'
+
+errors = []
+
+def matches(pattern, message)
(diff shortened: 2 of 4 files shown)


Gemfile.lock
```diff
@@ -230,4 +232,4 @@ RUBY VERSION
    ruby 2.4.1p111
 
 BUNDLED WITH
-   1.15.1
+   1.16.0.pre.1
```
colby-swandale: please don't bump the bundler version to a beta release.

arbonap:
Thanks @colby-swandale! I bundled with `1.15.4` now instead of with the prerelease.

spec/quality_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  files.each do |file|
+    next unless file.file?
+    relative_path = file.relative_path_from(Pathname('..').expand_path(__dir__))
+
+    skippable_directories = %w[source/man source/localizable source/layouts]
+
+    next if relative_path.to_s =~ Regexp.union(skippable_directories)
+    next if relative_path.to_s =~ %r{\Asource/v([\d.]+)} && Gem::Version.new($1) <= Gem::Version.new("1.15")
+    # this hackily skips blog posts touched before September 17th, 2017:
+    next if File.mtime(relative_path.to_s) < Time.parse("2017-09-17")
```
colby-swandale: This won't work as you expect because files from a fresh clone of the bundler-site repo will have a modified time of the current date-time, not when the file itself was created.

I noticed that each blog has a date in the filename, so maybe we can leverage that?

colby-swandale:
ping @arbonap are you still working on this PR?

esasse:
I’m closing this since there was no response from the author for almost two years.

PR #338, opened 2017-09-20
Update ruby and gems

hmistry:

None. Just maintenance update.

Update ruby and gems.


colby-swandale:
Thanks! @bundlerbot r+

PR #339, opened 2017-10-07
Remove Fastly env vars

indirect:
These are now provided via the Travis web API, and had to be updated
because the API key in question changed.


indirect:
@bundlerbot r+

PR #340, opened 2017-10-10
Fix link for the current getting started guide

abunashir:

In the homepage the `How Does Bundler Work?` link is pointing to the getting started guide from `v1.12` instead of using the current version.

This commit fixes this and it also replaces the hardcoded version number with the `current_version`, so in the future, we don't have to update it manually.


colby-swandale:
Thanks! 🙏  @bundlerbot r+

PR #341, opened 2017-10-10
Fix broken links for index page in v1.15

abunashir:

In the index page, there is a broken link and also two markdown style links which do not render those link properly.

This commit changes those to `haml` link helper and it also fixes the bundler setup guides path, which was changed in `v1.15`.


colby-swandale:
Thanks! @bundlerbot r+

PR #322, opened 2017-08-06
Rework guides listing

indirect:
This PR reworks the way we list guides, both on the main Docs page, and in the sidebar that shows anytime you're inside a guide.

Guides section on Docs page at `/docs`:

Sidebars on pages that are not command references:

PR #342, opened 2017-10-11
Fix broken links in the doc index page

abunashir:

In the `/v1.15` page, we have a section called `Use Bundler with`, where we have some links that were intended to point to the guides. But unfortunately, those links are broken and still seems to point to the old html file that does not exist anymore

This commit fixes those links and points those broken link to the correct guide path that changed in PR #322.


colby-swandale:
Thanks! @bundlerbot r+

PR #344, opened 2017-10-15
Minor updates on the deploying guide

abunashir:

In the `Deploying bundled applications` (Link: `/guides/deploying.html`) guide, there seems to be a broken link, the `executables` does not seems to be linking to the correct path. And, the guide also using a sample from the old Heroku's way of adding the configuration.

This commit fixes that link to the correct path, and it also updates the Heroku sample to their latest way of configuring environment variable.

Ref: https://devcenter.heroku.com/articles/bundler


colby-swandale:
Thanks @bundlerbot r+

Issue #254, opened 2016-09-14
Add the Bundler Issues page to the search


coilysiren:
ie. http://bundler.io/issues.html, as far as I can tell I can only get to it through google search

Ferdy89:
Hello @lynnco! I was taking a look about what was going on here and I noticed that `source/issues.html.md` doesn't have a [frontmatter](https://middlemanapp.com/basics/frontmatter/). I tested locally adding
```markdown
---
title: Issues
description: If you're running into an error, try reading the rest of this page for help.
---

# Bundler Issues
...
```
to `source/issues.html.md` and it worked. I'm happy to submit a PR for this if it makes sense.

<img width="502" alt="screen shot 2017-02-08 at 9 58 54 pm" src="https://cloud.githubusercontent.com/assets/1112217/22768769/d6cdb012-ee49-11e6-9405-c043fc1f14bc.png">

I also noticed other files with similar characteristics not showing, like `compatibility.html`, `conduct.html`, `older_versions.html` and others in the `/source` directory. These aren't listed as resources to be searched for in `config.rb`, maybe they should be?

deivid-rodriguez:
This page no longer exists since https://github.com/rubygems/rubygems/pull/4980, so this can be closed.

PR #345, opened 2017-10-16
Enable missing pages to be searchable

arbonap:

The problem was...
- Like @lynncyrin noted in Issue #254, not all pages come up in Bundler.io's search. 

My diagnosis was...
Specifically, the missing pages that should be searchable but currently weren't include:
- `compatibility.html`
- `conduct.html`
- `contributors.html`
- `older_versions.html`

My fix...
- I read through [the documentation for the Middleman search gem](https://github.com/manastech/middleman-search) that bundler.io uses and found out that in order to make a page searchable, we need to both give the page [frontmatter](https://middlemanapp.com/basics/frontmatter/) (the YAML at the top of the page), and append it to the `search.resources` array in `config.rb`. 
- I also deleted `'/doc/contributing/issues.html'` from the `search.resources` array since that page no long exists.
- I also added some CSS to give the search-box result elements a max-width. Otherwise, the width of these elements would be much wider than the search box itself if the page's description was long.
- To illustrate, this is what it would look like without giving the search box element a max-width:
- screen shot 2017-10-15 at 6 23 38 pm
- This is what it looks like with a max-width:
- screen shot 2017-10-15 at 6 22 35 pm

I chose this fix because...
- I think it's important for bundler users to be able to search for all bundler documentation pages.
- I think the CSS change will prevent search-results from looking too big. This will make the CSS more maintainable in the long-run. ✨

### DIFF
--- a/source/stylesheets/_search.scss
+++ b/source/stylesheets/_search.scss
@@ -49,7 +49,7 @@
 .search-list-ul {
   list-style-type: none;
   padding-left: 0;
-
+  
   hr {
     margin-top: 0;
     margin-bottom: 0;
--- a/config.rb
+++ b/config.rb
@@ -7,7 +7,7 @@ activate :syntax
 activate :i18n
 activate :sprockets
 activate :search do |search|
-  search.resources = ['index.html', "#{config[:current_version]}/", '/doc/contributing/issues.html']
+  search.resources = ['index.html', "#{config[:current_version]}/", 'compatibility.html', 'conduct.html', 'contributors.html', 'older_versions.html']
 
   search.index_path = 'search/lunr-index.json'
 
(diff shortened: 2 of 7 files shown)


colby-swandale:
I agree that the search results should have a max width but i think 280px is too small because it's making a lot of search results go below the fold on a 13" display. I find ~760px to be more suitable which it is un-officially is set to at the moment.

arbonap:
Sounds good @colby-swandale, I got rid of the CSS change 👌🏼

source/stylesheets/_search.scss
```diff
@@ -49,7 +49,6 @@
 .search-list-ul {
   list-style-type: none;
   padding-left: 0;
-
```
colby-swandale: I prefer not removing the space to keep with the current style.

arbonap: Sounds good! I added back in the space.

arbonap:
Thanks for the reviews @lynncyrin and @colby-swandale. Just added back in the missing whitespace in `_search.scss`. Let me know what y'all think.

colby-swandale:
Great! @bundlerbot r+

PR #346, opened 2017-10-18
Fix a broken link in the using group guide

abunashir:

In the `Using Group` guide, there seems to be a broken link,
Link: https://bundler.io/v1.15/guides/groups.html

This commit fixes that by changing the link to use `link_to_documentation` helper and hopefully in the future, it should work as expected, Thanks!


colby-swandale:
Thanks @bundlerbot r+

PR #347, opened 2017-10-24
Fix a broken link in using bundler guide

abunashir:
In the `Using Bundler In Applications` guide, the `git bisect` link seems to be broken.

This commit updates the link to the correct guide path.

Ref: https://bundler.io/v1.15/guides/using_bundler_in_applications.html#troubleshooting


colby-swandale:
Thanks! @bundlerbot r+

PR #348, opened 2017-11-01
Fix broken links to guides/groups.html

cowlibob:
Broken links during move from /groups.html to /guides/groups.html (between 1.14 and 1.15)


segiddins:
Thanks! @bundlerbot r+

PR #349, opened 2017-11-03
[Man] Ensure index.txt is built for 1.16+

segiddins:

The problem was the `.ronn` files need `index.txt` to be present to properly link things.

Closes https://github.com/bundler/bundler/pull/6143

My diagnosis was we needed to build `index.txt` for 1.16+

### DIFF
--- a/data/contributors.yml
+++ b/data/contributors.yml
@@ -1,1697 +1,1681 @@
 ---
-- :avatar_url: https://avatars1.githubusercontent.com/u/16457?v=3&s=40
+- :avatar_url: https://avatars2.githubusercontent.com/u/16457?v=4&s=40
   :commits: 445
   :href: https://github.com/hone
   :name: hone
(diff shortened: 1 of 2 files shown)


abunashir:
Amazing! that was really quick 🥇. I just tried to build the man pages locally and looks like the links are still broken and I think the issue now is with the references `id` in the `index.txt` file and the way we we are referencing it from any man pages. Here is the chunk from the generated `index.txt` file.

screenshot 2017-11-03 10 35 01

And here is how we are referencing from the man page (`bundler/man/bundle-config.ronn`), the second one is the `id` reference, format: `[text][id`.

screenshot 2017-11-03 10 39 00

I think we could update all the man pages references to the same pattern as generated one or maybe add support for both, what would you suggest? Thanks.

colby-swandale:
Thanks! @bundlerbot r+

PR #350, opened 2017-11-19
Edit out require directive for rack/cache in docs

arbonap:

The problem was...
- There were some outdated references in the Bundler rationale and setup guide docs that referred to how to use the require directive for `rack/cache`.

My diagnosis was...
- After reading the discussion on https://github.com/bundler/bundler-site/issues/130, I edited out the require directive for `rack/cache`since it's not needed anymore.

My fix...
- Edit out outdated documentation.

I chose this fix because...
- it's what was discussed in issue https://github.com/bundler/bundler-site/issues/130

### DIFF
--- a/source/v1.12/bundler_setup.html.md
+++ b/source/v1.12/bundler_setup.html.md
@@ -72,20 +72,7 @@ require 'rack-cache'
 require 'nokogiri'
 ~~~
 
-Astute readers will notice that the correct way to require the <code>rack-cache</code>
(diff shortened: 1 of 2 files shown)


arbonap:
I'm not entirely sure if these are the correct updates to the docs, but this is what I could gather from the discussion on issue https://github.com/bundler/bundler-site/issues/130. Happy to update my PR if it needs changes.

colby-swandale:
Thanks! @bundlerbot r+

PR #351, opened 2017-11-23
update bash shebang to '/usr/bin/env bash' form

jesperronn:
/bin/bash can be misleading for instance on macs using `brew install bash` and rvm, since it will point to the factory bash.

In general, scripts are considered more portable using  the shebang with `!#/usr/bin/env bash`.

More info: https://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash

May I also note that this ** PR is now consistent with all other script files in the repository** 👍 (I am a big fan of consistency)

### DIFF
--- a/source/v1.12/git_bisect.html.md
+++ b/source/v1.12/git_bisect.html.md
@@ -19,7 +19,7 @@ to test if they are not reset.
 Here's a minimal example script that runs the rake task `spec`:
 
 ~~~ bash
-#!/bin/bash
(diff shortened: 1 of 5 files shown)


colby-swandale:
Thanks! @bundlerbot r+

jesperronn:
Thanks to @colby-swandale (and also to @bundlerbot) for the fast response! 

(even robots deserve a "thank you" now and then)

PR #354, opened 2017-12-25
Add missing example for 'require' specification

sergiopantoja:
This commit adds an example that's consistent with this page:
https://github.com/bundler/bundler-site/blob/master/source/v1.16/rationale.html.haml

Previously, there was no example of what the documentation was describing.

If this PR looks good, I can update it to include updated documentation for other versions besides 1.16 also. Let me know!


colby-swandale:
Thanks! @bundlerbot r+

PR #357, opened 2018-01-18
update current core maintainers and fix overflow bug from contributor list

colby-swandale:
This PR just updates the core maintainer list to reflect the current team and also fixes a bug where the contributor list was overflowing.


indirect:
@colby-swandale thanks!

@bundlerbot r+

PR #359, opened 2018-01-18
Update core_team in the contributors import rake task to reflect the current core team

colby-swandale:
Update the `core_team` list to reflect the current team


colby-swandale:
@bundlerbot r+

PR #360, opened 2018-01-26
Rewrote page headers for most v.16 guides and added subcopy to Docs…

rubymorillo:
… main headers.

Given that the Bundler Docs site receives so much direct traffic, I wanted to try to add some sub-copy on the main Docs landing page to give more context around the types of content a user could find. Additionally, I wanted to standardize (most) of the Guide titles to make them easier to find when a user does a search for the issue.

To add two blurbs on the Docs main site, and to rewrite the titles of the v1.16 guides.

You can view the changes here:

- Docs landing page before the rewrite (what it currently looks like): 

- Docs landing page after the rewrite (notice the two blurbs under "Guides" and "Commands", as well as the revised Guide titles): 

I chose this fix because it's a quick fix and one that can translate into more Bundler users finding docs easier!

### DIFF
--- a/source/v1.16/guides/bundler_sharing.html.haml
+++ b/source/v1.16/guides/bundler_sharing.html.haml
@@ -1,9 +1,9 @@
 ---
-title: Sharing your code
+title: How to package and share code using Gemfile
 ---
 .container.guide
   %h2#sharing
-    Sharing your code
+    How to package and share code using Gemfile
   .contents
     .bullet
       .description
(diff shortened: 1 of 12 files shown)


source/v1.16/guides/bundler_sharing.html.haml
```diff
@@ -1,9 +1,9 @@
 ---
-title: Sharing your code
+title: How to package and share code using Gemfile
```
segiddins: `using a Gemfile`?

rubymorillo:
@segiddins good catch; updated!

indirect:
@bundlerbot r+

PR #366, opened 2018-02-21
Fix 404 to guides/git

ksolo:

When on the gemfile page, there is a link to learn more on using gems hosted via git. When clicked
I received a 404

The relative link reference was missing the `guide` in the path

I updated the link, and fixed one other small typo.


colby-swandale:
Thanks! @bundlerbot r+

PR #368, opened 2018-03-12
Update SSL steps to work on Windows

indirect:
These changes include a single-line command that will run the automated SSL troubleshooting script even on windows in a command prompt, and update the instructions for manually downloading and installing the latest rubygems to be a bit clearer.


indirect:
@bundlerbot r+

PR #369, opened 2018-03-15
Added Vagrant section & fixed jump links

rubymorillo:

The problem was...
A user reported a possible problem and fix for reconciling their rubygems certificate in vagrant (https://github.com/rubygems/rubygems/issues/1736#comment-373312394). We wanted to add this to the guide.

Additionally, while testing out the links in staging, I noticed that many of the jump links in the doc weren't properly set, so I made sure the anchor text matched the title of the section.

My diagnosis was...
1. add a possible solution for vagrant users
2. fix the jump links

My fix...
see above!

I chose this fix because...
it would help vagrant users access a possible solution && users would actually be able to jump around the doc successfully!

### DIFF
--- a/source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
+++ b/source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
@@ -27,6 +27,7 @@ possible, you can jump straight to [solutions for SSL issues][solutions-for-ssl-
     - [Automated SSL check][ssl-check]
     - [Updating Bundler][update-bundler]
     - [Updating RubyGems][update-rubygems]
+    - [Updating Time in Vagrant][update-vagrant]
     - [Updating CA certificates][updating-ca-certificates]
       - [Installing new RubyGems certificates][update-rubygems-certs]
       - [Installing new OS certificates][update-os-certs]
@@ -215,6 +216,10 @@ directory with RubyGems):
 
 Running `gem --version` should display the updated version.
 
+### Updating Time in Vagrant
+[update-vagrant]: #update-vagrant
+
+If the RubyGems certificate is newer than the time set inside of your Vagrant machine, you might have to set the correct time in Vagrant to resolve the error.
 
 ### Updating CA certificates
 [updating-ca-certificates]: #updating-ca-certificates
--- a/source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
+++ b/source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
@@ -219,9 +219,15 @@ Running `gem --version` should display the updated version.
 ### Updating System Clock
 [update-system-clock]: #updating-system-clock
 
-f your system clock is set to a time in the past or future, your machine will not be able to establish a secure connection to RubyGems.org. To resolve the issue, you will need to set your system clock to the current time.
+If your system clock is set to a time in the past or future, your machine will not be able to establish a secure connection to RubyGems.org. To resolve the issue, you will need to set your system clock to the current time. In a Linux VM, you can update the system clock by running `sudo ntpdate ntp.ubuntu.com`. 
 
-This issue most commonly shows up using Vagrant to run virtual machines on macOS. In a Linux VM, you can update the system clock by running sudo ntpdate ntp.ubuntu.com. For a more permanent fix, [check out this Stack Overflow answer about automating Vagrant clock updates](https://stackoverflow.com/questions/33939834/how-to-correct-system-clock-in-vagrant-automatically).
+Here are other possible solutions for updating a system clock:
+
+- macOS: [Set the time and date on your Mac](https://support.apple.com/kb/PH25523?locale=en_US)
+- Ubuntu: [Ubuntu Time Management](https://help.ubuntu.com/community/UbuntuTime)
+- Windows: [Getting Windows 10 time to sync with a time server](https://answers.microsoft.com/en-us/windows/forum/windows_10-other_settings-winpc/how-to-force-windows-10-time-to-synch-with-a-time/20f3b546-af38-42fb-a2d0-d4df13cc8f43)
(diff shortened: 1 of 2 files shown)


hmistry:
Edits look good to me. 👍 

A thought for future... I like to see guides that walk you through installations, how to do something, debug, etc be laid out in a step-by-step number list or equivalent format instead of paragraphs of journeys with multiple steps in a single paragraph. It's much easier to follow and helps in backtracking when you miss something.

source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
```diff
@@ -27,6 +27,7 @@ possible, you can jump straight to [solutions for SSL issues][solutions-for-ssl-
     - [Automated SSL check][ssl-check]
     - [Updating Bundler][update-bundler]
     - [Updating RubyGems][update-rubygems]
+    - [Updating Time in Vagrant][update-vagrant]
```
indirect: Let's change this title to `Updating System Clock`, since the clock being wrong will cause this problem on any machine, Vagrant or not.

source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
```diff
@@ -215,6 +216,10 @@ directory with RubyGems):
 
 Running `gem --version` should display the updated version.
 
+### Updating Time in Vagrant
+[update-vagrant]: #updating-time-in-vagrant
+
+If the RubyGems certificate is newer than the time set inside of your Vagrant machine, you might have to set the correct time in Vagrant to resolve the error.
```
indirect: Since this can apply to any machine, and not just Vagrant, how about something like this?

> If your system clock is set to a time in the past or future, your machine will not be able to establish a secure connection to RubyGems.org. To resolve the issue, you will need to set your system clock to the current time.
>
> This issue most commonly shows up using Vagrant to run virtual machines on macOS. In a Linux VM, you can update the system clock by running `sudo ntpdate ntp.ubuntu.com`. For a more permanent fix, check out [this stack overflow answer about automating Vagrant clock updates](https://stackoverflow.com/questions/33939834/how-to-correct-system-clock-in-vagrant-automatically).

rubymorillo:
@hmistry completely agree! Eventually I'd like to break this guide down even further—where the intro gets turned into a separate general article that explains the way, and the rest of the guide focuses on troubleshooting. But moving forward, we'll definitely adopt a step-by-step approach to our guides!

source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
```diff
@@ -215,6 +216,12 @@ directory with RubyGems):
 
 Running `gem --version` should display the updated version.
 
+### Updating System Clock
+[update-system-clock]: #updating-system-clock
+
+f your system clock is set to a time in the past or future, your machine will not be able to establish a secure connection to RubyGems.org. To resolve the issue, you will need to set your system clock to the current time.
```
colby-swandale: `If`

source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
```diff
@@ -215,6 +216,12 @@ directory with RubyGems):
...
+
+This issue most commonly shows up using Vagrant to run virtual machines on macOS. In a Linux VM, you can update the system clock by running sudo ntpdate ntp.ubuntu.com. For a more permanent fix, [check out this Stack Overflow answer about automating Vagrant clock updates](https://stackoverflow.com/questions/33939834/how-to-correct-system-clock-in-vagrant-automatically).
```
colby-swandale: I would remove all mentions of `Vagrant`, this problem can very easily happen on any system, not just virtual machines.

colby-swandale: > sudo ntpdate ntp.ubuntu.com.

The command needs the back ticks so it's highlighted. And i would mention that this command is only available on unix based operating systems.

colby-swandale: > https://stackoverflow.com/questions/33939834/how-to-correct-system-clock-in-vagrant-automatically

The marked answer in this post doesn't guarantee that it will permanently fix a clock that is out of sync, because it's only adjusting the time zone. I suggest using these links:

- https://support.apple.com/kb/PH25523?locale=en_US
- https://help.ubuntu.com/community/UbuntuTime
- https://answers.microsoft.com/en-us/windows/forum/windows_10-other_settings-winpc/how-to-force-windows-10-time-to-synch-with-a-time/20f3b546-af38-42fb-a2d0-d4df13cc8f43 (There doesn't seem to be a knowledge doc about this 😢 )

rubymorillo:
@colby-swandale thanks for catching the typo, and good call on the additional resources. I've removed mentions of Vagrant and created a bulleted list prefaced as "other _possible_ solutions for updating a system clock", linking out to each of the resources above (https://github.com/bundler/bundler-site/pull/369/commits/372a18a9be261cd7c81b202dfa65e2274ad81491)

source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
```diff
@@ -215,6 +216,18 @@ directory with RubyGems):
 
 Running `gem --version` should display the updated version.
 
+### Updating System Clock
+[update-system-clock]: #updating-system-clock
+
+If your system clock is set to a time in the past or future, your machine will not be able to establish a secure connection to RubyGems.org. To resolve the issue, you will need to set your system clock to the current time. In a Linux VM, you can update the system clock by running `sudo ntpdate ntp.ubuntu.com`.
```
hmistry: We can be more universal by saying `In Linux,...` instead of `In a Linux VM, ...`. The command runs on any Linux machine whether it's the native OS on a computer or a Linux VM on Mac/Windows on a computer.

indirect:
@bundlerbot r+

PR #370, opened 2018-03-17
Replace smart quotes with quotes

olleolleolle:

The example code was not copy-and-pastable due to smart quotes.

My diagnosis was that the smart quotes would be syntax errors in Ruby.

My fix was the replace the smart quote marks in the Markdown file.

It seemed easiest.


segiddins:
Thanks!
@bundlerbot r+

PR #371, opened 2018-03-17
Add meta tag for specifying language

shime:

Chrome was suggesting to translate pages from Danish.

The meta tag for specifying language was missing.

I added the missing meta tag.

I'm not aware about any other possible options.

### DIFF
--- a/source/layouts/base.haml
+++ b/source/layouts/base.haml
@@ -3,6 +3,7 @@
   %head
     %meta(charset="utf-8")
     %meta(http-equiv="X-UA-Compatible" content="IE=edge")
+    %meta(http-equiv="Content-Language" content="en")
     %meta(name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no")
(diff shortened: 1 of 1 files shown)


segiddins:
This probably needs to handle the Spanish pages we have translated?

hmistry:
@segiddins true and polish I believe. Good catch.

esasse:
Ping @shime do you still want to do this? Thanks.

source/layouts/base.haml
```diff
@@ -3,6 +3,7 @@
   %head
     %meta(charset="utf-8")
     %meta(http-equiv="X-UA-Compatible" content="IE=edge")
+    %meta(http-equiv="Content-Language" content="en")
```
olleolleolle, line 6: Riffing on how that could be solved:

Could that be calculated like something such as `current_page.data.locale || "en"`, and then the pages that _are_ translated, they could get a locale annotation, either directly in their YAML preamble, or somehow calculated from their filename? Example of a translated file: https://github.com/rubygems/bundler-site/blob/master/source/localizable/v1.15/guides/bundler_setup.es.html.haml

olleolleolle (commented):
This PR made me aware of the cool translations that have been done.

PR #373, opened 2018-04-07
created march update post

rubymorillo:
Each month, we're cross-posting monthly updates from Ruby Together on the Bundler blog. I created the new post, based on the updates already published in Ruby Together.

### DIFF
--- a/source/conduct.html.md
+++ b/source/conduct.html.md
@@ -1,47 +1,136 @@
----
-title: Code of conduct
-description: Rules to adhere to while contributing to Bundler
----
-
 # Bundler Code of Conduct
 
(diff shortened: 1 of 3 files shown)


source/conduct.html.md
```diff
@@ -1,47 +1,136 @@
----
-title: Code of conduct
```
segiddins: do we no longer need the YAML front matter?

indirect: maybe the automatic generator didn’t catch the file for some reason? I’m not sure.

indirect:
...and [this](https://travis-ci.org/bundler/bundler-site/builds/363672996?utm_source=github_status&utm_medium=notification) is why I should never use the merge button. blahhhhhh

rubymorillo:
@indirect let me know if I need to change anything else!

Issue #372, opened 2018-04-06
Add documentation for `glob` to "How to install gems from git repositories" page


tcdowney:
Hello!

Recently I needed to install a gem from a Git repository, but the gem was nested within a subdirectory of the repository. Looking at the [existing docs](http://bundler.io/guides/git.html) I couldn't find anything that would allow me to do this. I eventually found an issue that mentioned the `glob` parameter and it worked, but it would be nice if this was documented more explicitly.

For example, in order to install [this gem](https://github.com/cloudfoundry/copilot/tree/master/sdk/ruby) from its Git source, I can use `glob` like this:
```
gem 'cf-copilot', git: 'https://github.com/cloudfoundry/copilot.git', branch: 'master', glob: 'sdk/ruby/*.gemspec'
```

Thanks!

rubymorillo:
thanks @tcdowney! would you like to open up a PR in the doc to have that added in? https://github.com/bundler/bundler-site/blob/master/source/v1.16/guides/git.html.haml

tcdowney:
Thanks for merging the PR!

PR #377, opened 2018-05-05
Document git :glob option for installing gems from subdirectories

tcdowney:
I initially logged Issue #372 for this, so just copying what part of what I said in there:

> Recently I needed to install a gem from a Git repository, but the gem was nested within a subdirectory of the repository. Looking at the [existing docs](http://bundler.io/guides/git.html) I couldn't find anything that would allow me to do this. I eventually found an issue that mentioned the `glob` parameter and it worked, but it would be nice if this was documented more explicitly.

Just an option that's missing from the existing git installation documentation.

Added a short blurb to the documentation to make this option more discoverable for folks. 😄 

Not sure how you all choose what gems to use as examples. I just put down the one I needed to discover to use `:glob` for to install.

Open to any and all suggestions to clarify the copy or choose a different gem as an example.

Thanks!

### DIFF
--- a/source/v1.16/guides/git.html.haml
+++ b/source/v1.16/guides/git.html.haml
@@ -33,6 +33,14 @@ title: How to install gems from git repositories
         gem 'nokogiri', '1.7.0.1', :git => 'https://github.com/sparklemotion/nokogiri'
     .bullet
       .description
+        If the gem is located within a subdirectory of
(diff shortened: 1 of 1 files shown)


rubymorillo:
this looks good to me! cc @indirect @colby-swandale

segiddins:
@bundlerbot r+

PR #382, opened 2018-06-15
Remove gramatical error

Samsinite:

The problem was a grammar error

My diagnosis was reading the docs and seeing the grammar error.

My fix was to fix the grammar error.

I chose this fix because I read these docs from a link inside of ruby weekly.


colby-swandale:
Thanks! @bundlerbot r+

PR #384, opened 2018-06-21
Fixes Broken links

### LINKED ISSUE #383: Broken link in "how to create a bundler plugin" docs
Broken link under "Getting started with development -> 1. Create a gem".

Link text is: "Create a gem using this guide" and points to https://bundler.io/v1.16/guides/guides/creating_gem.html

jasonkarns:

The problem was... broken links on the bundler website

My diagnosis was... mostly incorrect relative URLs.

My fix... use the correct URLs

I chose this fix because... keeps the links working.

fixes #383

### DIFF
--- a/source/blog/2016-04-28-the-new-index-format-fastly-and-bundler-1-12.html.md
+++ b/source/blog/2016-04-28-the-new-index-format-fastly-and-bundler-1-12.html.md
@@ -33,4 +33,4 @@ We added a bunch of other small tweaks, features, and bugfixes, so be sure to ch
 
 ## Updating
 
-To get the newest version of Bundler, run `gem install bundler`. If you have any issues, please check out our [issues guide](https://github.com/bundler/bundler/blob/master/ISSUES.md) and let us know!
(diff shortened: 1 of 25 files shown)


jasonkarns:
All in all the link checker found 107 broken links. This PR fixes maybe half of them (it does not fix any links from the man pages, where the majority of the broken links were found). Some things of special note:

- git-bisect guide doesn't exist at all in versions 1.12, 1.13, or 1.14 so in these versions, links point to the versionless  "top level" git-bisect guide (2b85913)
- https://bundler.io/gemfile_ruby.html links to https://bundler.io/bundle_platform.html and works under v1.16 but not without version.
- https://bundler.io/v1.16/bundle_config.html (also for versions 1.12, 1.13, 1.14, and 1.15) has a bunch of broken links to manpages but unsure where the source files are
- https://bundler.io/v1.16/bundle_platform.html has broken links but unsure of source file

segiddins:
Thanks for this!

@bundlerbot r+

PR #392, opened 2018-08-15
Update sprockets and nokogiri to fix security vulnerabilities

greysteil:
Given that this repo is only used to generate a static site we're not exposed to the vulnerabilities in these gems. Doesn't hurt to patch them anyway, though.


colby-swandale:
Thanks!

PR #394, opened 2018-08-16
Fixed the example of using Bundler in a single-file Ruby script

savonarola:

The example of using Bundler in a single-file Ruby script https://bundler.io/v1.16/guides/bundler_in_a_single_file_ruby_script.html fails with 

```
>ruby bundler_inline_example.rb
Gems installed and loaded!
Traceback (most recent call last):
bundler_inline_example.rb:11:in `<main>': uninitialized constant Nap (NameError)
```

`nap` gem has no `Nap` namespace currently. It has its `VERSION` constant in `REST` module.

Use the correct namespace. 

¯\\_(ツ)_/¯


greysteil:
❤️

PR #395, opened 2018-08-18
Created "how to use Bundler with Docker guide"

### LINKED ISSUE #386: Add a page discussing ways to use Bundler with Docker
Based on bundler/bundler#6524, it looks like the Docker default images for Ruby have cleverly engineered a situation where only one bundle and one application can exist inside the docker container.

For users who want to have more than one Gemfile, or simply install gems via RubyGems and use them as system gems, this situation is confusing, and has historically led to a lot of bug reports and even pull request against Bundler.

Instead, we want to document how Bundler handles binstubs, document how the default Docker Ruby image handles binstubs, and then document how the user can configure Bundler within Docker to support whatever setup it is that they want.

More details about those situations can be found in the comments on the ticket linked above.

rubymorillo:

closes #386 

The problem was...

Per #386:

> Based on bundler/bundler#6524, it looks like the Docker default images for Ruby have cleverly engineered a situation where only one bundle and one application can exist inside the docker container.

> For users who want to have more than one Gemfile, or simply install gems via RubyGems and use them as system gems, this situation is confusing, and has historically led to a lot of bug reports and even pull request against Bundler.

> Instead, we want to document how Bundler handles binstubs, document how the default Docker Ruby image handles binstubs, and then document how the user can configure Bundler within Docker to support whatever setup it is that they want.

...there

My diagnosis was...

...to create a guide based on the fix @eanlain outlined [in this comment](https://github.com/bundler/bundler/pull/6524#issuecomment-406726953)

My fix...

...to create a guide based on the fix one of our users outlined [here](https://github.com/bundler/bundler/pull/6524#issuecomment-406726953)

cc @indirect

### DIFF
--- a/source/v1.16/guides/bundler_docker_guide.html.md
+++ b/source/v1.16/guides/bundler_docker_guide.html.md
@@ -1,0 +1,31 @@
+---
+title: How to use Bundler with Docker
+---
+
+# How to use Bundler with Docker
+
+## Introduction 
+
+If you have tried to bundle an application in Docker with Bundler v1.16 or earlier, you may run into issues with explicitly setting `GEM_HOME`, `BUNDLE_BIN`, `BUNDLE_PATH`, and with adding the `BUNDLE_BIN` to the `PATH` in your Dockerfile similar to this:
+
+```
+ENV GEM_HOME="/usr/local/bundle"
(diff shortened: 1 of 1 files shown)


indirect:
@bundlerbot r+

PR #401, opened 2018-09-15
Switch groups guide to standard hash syntax

abraham:
Update to hash syntax pretty much everyone is using these days.


greysteil:
Also reasonable!

PR #402, opened 2018-09-15
Switch setup guide to standard hash syntax

abraham:
Update to standard hash syntax pretty much everyone is using.


greysteil:
👍 👍 

Thanks for these @abraham

PR #406, opened 2018-10-14
Clarify a sentence in rationale.html.haml

fdmarcin:

The code comment could previously suggest that `bundle` is a shortcut of `bundle install` **within the command** `bundle install`.

It can be confusing to the user.

Clarify that `bundle` on its own is a shortcut of `bundle install`. Of course, please correct me if I'm wrong.
Also, let me know if the proposed code comment is too long for the snippet, and should be broken down into two lines or rephrased.

### DIFF
--- a/source/v1.16/rationale.html.haml
+++ b/source/v1.16/rationale.html.haml
@@ -37,7 +37,7 @@
       After declaring your first set of dependencies, you tell bundler to go get them:
 
     :code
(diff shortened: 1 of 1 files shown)


colby-swandale:
Nice find! Thank you for the PR.

@bundlerbot r+

colby-swandale:
Sorry this took so long getting in 🙏

colby-swandale:
@bundlerbot r+

PR #409, opened 2018-10-23
Rackup 'command not found' error for sinatra

ParamagicDev:

I tried following the guide for Sinatra apps and when I would run:
$ bundle exec rackup
I'd receive an error that the command did not exist.

Because I had not run
$ bundle install --binstubs
previously in my app, I did not have a bin folder to exec gem commands like
$ bundle exec rackup
in this case

Fixed documentation to reflect that the bundle install must be run w/ --binstubs flag to ensure a folder of gem executables is created

I chose this fix because I was unable to get
$ bundle exec rackup
To work on my server but
$ ruby app/sinatra_app.rb would work just fine.

I also chose to fix this because I'm a new web developer with little experience looking to get my feet wet contributing to open source and had not found any decent solutions when googling the problem.

### DIFF
--- a/source/v1.16/guides/sinatra.html.haml
+++ b/source/v1.16/guides/sinatra.html.haml
@@ -46,7 +46,3 @@ title: How to use Bundler with Sinatra
         $ bundle exec rackup
       .description
         The code should no longer produce the error and your development server will be up and running
-        
-        
-
-        
--- a/source/v1.16/guides/sinatra.html.haml
+++ b/source/v1.16/guides/sinatra.html.haml
@@ -30,3 +30,23 @@ title: How to use Bundler with Sinatra
         Start your development server with rackup, and Sinatra will be loaded via Bundler.
       :code
         $ bundle exec rackup
+      
+    .bullet  
+      .description
+        If you run the above command and receive a 'command not found' error
+        Ensure you have a bin folder in your projects root directory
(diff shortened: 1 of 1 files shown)


source/v1.16/guides/sinatra.html.haml
```diff
@@ -30,3 +30,19 @@ title: How to use Bundler with Sinatra
...
+    .bullet  
+      .description
+        If you run the above command and receive a 'command not found' error
+        Ensure you have a bin folder in your projects root directory
+        Check to see that there is a file named 'rackup' within that folder
+        If you notice the file is not there run the following command:
+      :code
+        $ bundle install --binstubs
+      .description
+        This will install the executables needed for bundle exec
```
colby-swandale: `bundle exec` does not need to the executables in the `bin` directory to work. They exist so that you don't have to run `bundle exec <command>`.

What i suspect happened is that you didn't run `bundle install` after you created the Gemfile. We should instead append the tutorial to include running `bundle install` after creating the `Gemfile`. (Line 14)

ParamagicDev: You appear to be right. I had run bundle install multiple times however and the only thing that fixed it for me was running bundle install --binstubs. 

However, after SSHing today and removing the bin directory it appears to still work so I'm not sure what was going on.

PR #413, opened 2018-11-16
update rack to the latest version

colby-swandale:

The current version of rack contains a few reported vulnerabilities.

Update rack to the latest version


colby-swandale:
@bundlerbot r+

PR #434, opened 2019-01-04
fix note about Bundler 2 lockfile being formatted slightly different

colby-swandale:
There is a paragraph in the Bundler 2 upgrade that talks about the Gemfile.lock formatting being changed. This change was pushed back to Bundler 3.


colby-swandale:
@bundlerbot r+

PR #436, opened 2019-01-07
Change RubyGems requirement to v2.5.0 from v3.0.0 at minimum

0gajun:
The problem was that "How to upgrade to Bundler 2" document made users to misunderstand RubyGems requirement at minimum. 

Indeed, in bundler v2.0.0, the requirement was v3.0.0 at minimum.
However, since bundler v2.0.1, the requirement was relaxed to v2.5.0 from v3.0.0.

ref : https://bundler.io/blog/2019/01/04/an-update-on-the-bundler-2-release.html

None

My fix is changing RubyGems requirement to v2.5.0 from v3.0.0 in "How to Upgrade to Bundler 2" document.

None


colby-swandale:
Thanks for picking up on that!

colby-swandale:
@bundlerbot r+

PR #431, opened 2019-01-03
Fixed two typos in Bundler 2 release post

radar:
I've removed a small typo in the Heroku buildpack section.

I've also corrected the link to the issues guide at the bottom of the post.

PR #440, opened 2019-01-15
Make Heroku upgrade info match blog post

dgollahon:
The guide on the website has a hanging "by running:" followed by nothing which is confusing. Someone already noticed and fixed this in #431 but those didn't get translated to the upgrade guide on the website. I've copied the entire question wording verbatim for consistency, but if someone thinks it'd be better not to include the note about support, question change, etc. I can update this. At the very least, the "by running" should be removed so the sentence makes sense.

The problem was...

My diagnosis was...

My fix...

I chose this fix because...


indirect:
Thanks!

@bundlerbot r+

PR #447, opened 2019-03-04
Fix 'BUNDLED_WITH' -> 'BUNDLED WITH'

alyssais:

The problem was that the section name `BUNDLED_WITH` isn't actually a section in a Bundler lockfile.

My diagnosis was that this should have said `BUNDLED WITH`.

My fix was to change `BUNDLED_WITH` to `BUNDLED WITH`.

I chose this fix because it's clearly what was intended.


segiddins:
Thanks!
@bundlerbot r+

PR #450, opened 2019-03-11
Heroku now supports Bundler 2

danarnold:
Heroku now supports Bundler 2 with the official Ruby buildpack, so the documentation can now reflect this.


colby-swandale:
Thanks

@bundlerbot r+

PR #456, opened 2019-04-02
Remove useless cat

sikevux:

The problem was...

*  That there's useless cats in the documentation

My diagnosis was...

* That it's a semantics change that would make the documentation look better

My fix...

* Is to remove the cats and only use grep

I chose this fix because...

* It's the most common way to solve cat pipes to grep


simi:
nooo

esasse:
@bundlerbot r+

PR #460, opened 2019-05-10
Add security.txt, link to HackerOne

indirect:
The `security.txt` file is RFC-based standard for how to report security issues, and we should also link to the HackerOne site from the actual webpage.


indirect:
@bundlerbot r+

.ruby-version
```diff
@@ -1 +1 @@
-2.5.1
+2.6.3
```
segiddins: travis also needs to be updated?

indirect: whoops, yes it does!

indirect:
@bundlerbot r=segiddins

PR #466, opened 2019-05-28
Update bundler_2_upgrade.html.md

jason-rutherford:
I _think_ this PR is correct because I believe Bundler 2 requires RubyGems 3.0.0 or higher.

The Upgrade Guide to Bundler 2 appears to say that RubyGems 2.5.0 or higher is required. I don't think that is correct. I think it should say RubyGems 3.0.0 or higher.

n/a

Update documentation to say Bundler 2 requires at least RubyGems 3.0.0.

Both of these pages current state that bundler 2.0 "Removed support for RubyGems < 3.0.0"

* https://bundler.io/whats_new.html
* https://bundler.io/blog/2019/01/03/announcing-bundler-2.html


esasse:
Thanks, but Bundler 2.0.1 relaxed the requirement to 2.5.0: https://bundler.io/blog/2019/01/04/an-update-on-the-bundler-2-release.html

PR #472, opened 2019-07-29
Fix HackerOne link

halorgium:

The problem was a broken link. 

My diagnosis was a typo in the source. 

My fix is a `,`. 🎉 

I chose this fix because it was easy and fixes the problem.


hsbt:
@bundlerbot r+

PR #477, opened 2019-08-14
Documentation page for Bundler metric reporting and collecting functionality

Akrabut:
**This work has been done as part of Google Summer of Code 2019**
This is my suggested documentation for the metric collecting and reporting functionality [that will be added to Bundler](https://github.com/bundler/bundler/pull/7298).

I wasn't sure where to place it in the layout so I'll await your feedback or let you place it.

@indirect

### DIFF
--- a/source/v2.0/guides/metrics.html.md
+++ b/source/v2.0/guides/metrics.html.md
@@ -1,0 +1,60 @@
+
+# Bundler Metrics
+
+Recently, a metric collection and reporting system for Bundler [has been introduced](https://github.com/bundler/bundler/pull/7298).
+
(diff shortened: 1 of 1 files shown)


simi:
If I understand it well, this feature was never finished as originally intended and it was later replaced by https://ecosystem.rubytogether.org/. I suggest to close this case. It would be still nice to share some details and links at bundler.io to this project, but this is out of scope IMHO and it would be easier to start from scratch.

simi:
@indirect @deivid-rodriguez ok to close?

indirect:
Yes, I think so. 👍🏻

PR #478, opened 2019-08-29
Fix blog command syntax highlighting

abraham:
Readers might incorrectly copy the command and try to execute `run`.


abraham:
@esasse ping

esasse:
@bundlerbot r+

PR #483, opened 2019-10-23
Remove extra 'the' article

ritaritual:

Some extra 'the' article

Just removed it


esasse:
@bundlerbot r+

esasse:
Thanks @ritaritual !

PR #487, opened 2019-11-18
Clarify and deepen the discussion of Gemfile.lock in a gem repository

ianfixes:

The guidance on checking a `Gemfile.lock` into source control (for a gem library) has reversed in the past few years, but that's not due to a change in how the `Gemfile.lock` actually works.  So old articles about best practices (https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/) and new articles (https://bundler.io/v2.0/guides/faq.html#committing-lockfiles-in-libraries) are both technically correct.  

The documentation on the Bundler site does a poor job of conveying the nuance of the issue.  This was discussed here:
https://github.com/bundler/bundler/issues/5879#issuecomment-551967102

This patch updates the documentation from saying "do this one thing" to explaining the pros and cons of committing a Gemfile.lock to source control.  It also does a better job of explaining the problem to be solved: the tendency of new versions of dependencies to break `bundle install` independently of changes to the actual gem.

I chose this fix because what works for open-source gems may not be best for privately-maintained gems, and the tooling that each project uses in its development workflow may be different.

### DIFF
--- a/source/v2.0/guides/faq.html.haml
+++ b/source/v2.0/guides/faq.html.haml
@@ -219,6 +219,6 @@ title: Frequently Asked Questions
           <strong>A</strong>: To avoid the tradeoff between breaking `gem install` and breaking `bundle install`,
           one solution is to simply test both.  By keeping the `Gemfile.lock` in source control, it's possible to
           `bundle install` with and without the lockfile (by deleting it).
-  
+        %p
           To go a step further, running daily tests of `bundle install` without the lockfile is the most proactive
           solution to keeping the dependencies in working order.  This the approach taken by tools like Dependabot.
--- a/source/v2.0/guides/faq.html.haml
+++ b/source/v2.0/guides/faq.html.haml
@@ -181,12 +181,44 @@ title: Frequently Asked Questions
           <strong>Q</strong>: Should I commit my `Gemfile.lock` when writing a gem?
 
         %p
-          <strong>A</strong>: Yes. When Bundler first shipped, the
-          `Gemfile.lock` was gitignored inside gems.  Over time, however, it
-          became clear that this practice forces the pain of broken dependencies
-          onto new contributors, while leaving existing contributors potentially
-          unaware of the problem. Since `bundle install` is usually the first
-          step towards a contribution, the pain of broken dependencies would
-          discourage new contributors from contributing. As a result, we have
-          revised our guidance for gem authors to now recommend checking in the
-          lock for gems.
+          <strong>A</strong>: The answer depends on how you prefer to balance the experiences of your gem's
+          developers and its users. Broken dependencies occur externally to the development of a gem -- 
+          they can break even when the gem itself remains unchanged, typically because a new version of a 
+          dependency may fail to work with the gem.  Committing `Gemfile.lock` (or not)
+          will determine whether developer or users will be the first to discover these broken dependencies.
+          
+        %p
+          <strong>Q</strong>: What are the implications of committing my `Gemfile.lock` when writing a gem?
+
+        %p
+          <strong>A</strong>: The presence of a `Gemfile.lock` in a gem's repository ensures that 
+          a fresh checkout of the repository (e.g. by a new contributor) will complete its `bundle install`
+          successfully.  
+        %p
+          The disadvantage to committing `Gemfile.lock` means that it is the <i>users</i> of your gem who 
+          will be the first affected by broken dependencies -- their `gem install` may fail.
+        %p
+          When Bundler first shipped, the `Gemfile.lock` was `.gitignore`d inside gems.  For open source 
+          projects, this meant that `bundle install` could work when code was tested and merged, yet fail 
+          at a later point -- often when a new contributor was running `bundle install` for the first time;
+          the pain of these broken dependencies would tend to discourage them from contributing.  This by 
+          itself was significant enough that we've revised our guidance for gem authors to now recommend 
+          checking in the lock for gems.
+        %p
+          For private gems (implying a closed team in which the developers are all "maintainers"), having
+          broken dependencies in a development environment may be entirely appropriate -- a
+          `bundle install` would be initiated (either locally or as part of a build process) by someone 
+          prepared to fix any problems that arise.  In other words, the most proactive way to alert 
+          developers to broken dependencies is to remove the `Gemfile.lock` from the repository.
+          
+        %p
+          <strong>Q</strong>: In terms of Continuous Integration, what are the best practices for maintaining 
+          a repository that contains a gem?
+
+        %p
+          <strong>A</strong>: To avoid the tradeoff between breaking `gem install` and breaking `bundle install`,
+          one solution is to simply test both.  By keeping the `Gemfile.lock` in source control, it's possible to
+          `bundle install` with and without the lockfile (by deleting it).
+  
+          To go a step further, running daily tests of `bundle install` without the lockfile is the most proactive
+          solution to keeping the dependencies in working order.  This the approach taken by tools like Dependabot.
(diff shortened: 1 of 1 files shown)


source/v2.0/guides/faq.html.haml
```diff
@@ -181,12 +181,44 @@ title: Frequently Asked Questions
...
-          unaware of the problem. Since `bundle install` is usually the first
-          step towards a contribution, the pain of broken dependencies would
-          discourage new contributors from contributing. As a result, we have
-          revised our guidance for gem authors to now recommend checking in the
-          lock for gems.
+          <strong>A</strong>: The answer depends on how you prefer to balance the experiences of your gem's
+          developers and its users. Broken dependencies occur externally to the development of a gem -- 
+          they can break even when the gem itself remains unchanged, typically because a new version of a 
+          dependency may fail to work with the gem.  Committing `Gemfile.lock` (or not)
+          will determine whether developer or users will be the first to discover these broken dependencies.
```
deivid-rodriguez: I disagree with the last sentence. Whether you commit the lockfile or not is independent from who will discover broken dependencies.

You can commit your lockfile, and still find breakage very early (for example, because dependabot generated an automated PR in real time after the release of the broken dependency), and reversely, you can gitignore your lockfile and still not find out about a broken dependency until your users tell you (for example, because your repo does not have active CI builds or new PRs).

source/v2.0/guides/faq.html.haml
```diff
@@ -181,12 +181,44 @@ title: Frequently Asked Questions
...
+        %p
+          <strong>Q</strong>: What are the implications of committing my `Gemfile.lock` when writing a gem?
+
+        %p
+          <strong>A</strong>: The presence of a `Gemfile.lock` in a gem's repository ensures that 
+          a fresh checkout of the repository (e.g. by a new contributor) will complete its `bundle install`
+          successfully.  
+        %p
+          The disadvantage to committing `Gemfile.lock` means that it is the <i>users</i> of your gem who 
+          will be the first affected by broken dependencies -- their `gem install` may fail.
```
deivid-rodriguez: Ditto, gitignoring your lockfile does not guarantee this.

source/v2.0/guides/faq.html.haml
```diff
@@ -181,12 +181,44 @@ title: Frequently Asked Questions
...
+          at a later point -- often when a new contributor was running `bundle install` for the first time;
+          the pain of these broken dependencies would tend to discourage them from contributing.  This by 
+          itself was significant enough that we've revised our guidance for gem authors to now recommend 
+          checking in the lock for gems.
+        %p
+          For private gems (implying a closed team in which the developers are all "maintainers"), having
+          broken dependencies in a development environment may be entirely appropriate -- a
+          `bundle install` would be initiated (either locally or as part of a build process) by someone 
+          prepared to fix any problems that arise.  In other words, the most proactive way to alert 
+          developers to broken dependencies is to remove the `Gemfile.lock` from the repository.
```
deivid-rodriguez: I'm good with including something like this :+1:.

source/v2.0/guides/faq.html.haml
```diff
@@ -181,12 +181,44 @@ title: Frequently Asked Questions
...
+          <strong>Q</strong>: In terms of Continuous Integration, what are the best practices for maintaining 
+          a repository that contains a gem?
+
+        %p
+          <strong>A</strong>: To avoid the tradeoff between breaking `gem install` and breaking `bundle install`,
+          one solution is to simply test both.  By keeping the `Gemfile.lock` in source control, it's possible to
+          `bundle install` with and without the lockfile (by deleting it).
+        %p
+          To go a step further, running daily tests of `bundle install` without the lockfile is the most proactive
+          solution to keeping the dependencies in working order.  This the approach taken by tools like Dependabot.
```
deivid-rodriguez: Not sure if this is what you mean here by "running daily tests", but I think if you include an entry in your CI to test your application or library against, you should do it on an scheduled CI, not on regular PRs. The rationale is the same. Contributors can get randomly failing CIs just because of a release of a dependency, and that's again putting burden of broken dependencies onto new contributors (they might not be able to tell that the problem is not their fault, they might get discouraged by being unable to get CI green, etc).

I don't think this is exactly dependabot's approach though. I mean, I'm not fully sure how it works internally, but if you're committing your lockfile, I assume it's running something like `bundle update <gem>` with specific dependencies and PRs with the updated lockfiles. I don't think this is "running tests without the lockfile", so we should clarify it.

ianfixes:
How attached are you to the original text and/or the spirit of the original text?  I tried to leave it in there, but as you can see there's a bit of a gap between the present "we recommend checking in Gemfile.lock" and my proposed "here are the challenges in maintaining a gem library".

I could probably write this a bit more clearly if you don't mind me scrapping what's there and rewriting from the perspective of the actual problem (broken dependencies) being solved.

source/v2.0/guides/faq.html.haml
ianfixes: What I'm getting at here is that by committing the Gemfile.lock, you guarantee that it will always work for developers who do a fresh checkout of the code.  So in that case, developers are shielded from the problem and users aren't.  I can word this a bit better.

ianfixes:
I think I've properly removed the incorrect statements and also put more detail into "ideal" CI configuration (PR builds with Gemfile.lock, scheduled builds without).  Does this look OK?

source/v2.0/guides/faq.html.haml
```diff
@@ -181,12 +181,48 @@ title: Frequently Asked Questions
...
+          The consequence of committing `Gemfile.lock` is that the developers -- those in a position to 
+          resolve broken dependencies -- may be shielded from seeing the issue in the first place.  This
+          can be either good or bad.
+        %p
+          When Bundler first shipped, the `Gemfile.lock` was `.gitignore`d inside gems.  For open source 
+          projects, this meant that `bundle install` could work when code was tested and merged, yet fail 
+          at a later point -- often when a new contributor was running `bundle install` for the first time;
+          the pain of these broken dependencies would tend to discourage them from contributing.  This by 
+          itself was significant enough that we've revised our guidance for gem authors to now recommend 
+          checking in the lock for gems.
```
AlexWayfer: 1. "A new contributor" can fix dependencies issues manually (temporary, and then create PR, for example).
2. This is a problem of versions-lock in `Gemfile` (or `gemspec`), like a bug in software, and it should be detected and fixed.

ianfixes:
It sounds like there are a few separate concerns at play here.  

The spirit of the existing documentation with respect to versions and dependencies seems to favor compatibility -- being flexible in what versions you will accept makes your gem better able to resolve dependencies when used alongside other gems in the same project.

What you're suggesting by locking versions in the `.gemspec` seems to favor keeping the individual gem in a guaranteed-working state, at the expense of the flexibility described above.

You're correct that a new contributor is in a position to fix the dependencies,  in the sense that they have the code in front of them in an editor (or likely the command line).  But my conversations with the developers on this topic (in other threads) is that new contributors are less likely to contribute if the project arrives on their system in a broken state.  

I am not contributing this patch to dispute that experience; I am trying to elaborate on the available design choices and their implications/limitations.  @deivid-rodriguez what would you prefer I do here?

deivid-rodriguez:
Hi @ianfixes! I'm sorry I dropped the ball here.

So I re-read this PR again, and it felt like I was going to keep leaving feedback forever just to end up describing bundler's team opinionated vision about this. So, instead, I opened #501 to do that, but still try to give more information so that people can more informedly disagree :)

I also included a separate question about how `Gemfile` and `Gemfile.lock` are used inside gems to try to clarify some other stuff that came up here.

So, if you are ok, I'll close this PR and go with #501, but by no means I intend to dismiss your contribution, it started an interesting discussion.

ianfixes:
I can't tell you how angry I am to hear that the official docs will be written by an expert instead of a fairly ignorant newcomer like myself 😉 

I'm glad I got the discussion started, it's been difficult for quite some time to get clarity on this topic.  I'll add some comments to #501

deivid-rodriguez:
Maybe an expert in other matters, but definitely a bad prose writer, so feedback is super welcome!

PR #493, opened 2019-12-11
Add GitHub Actions workflow

esasse:

We want to automate merge of dependabot PRs.

We need an active check in the repository so that dependabot can automatically merge its PRs after all checks are passed.

I've added a GitHub Action workflow that builds the site.

It was a [suggestion from @indirect](https://bundler.slack.com/archives/C0ZRUQ78V/p1575591631001700) and I agree it's the best option.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,26 @@
+name: Middleman build
+on: push
+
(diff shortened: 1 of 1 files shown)


indirect:
@bundlerbot r+

PR #495, opened 2019-12-14
Remove extra Comma; Remove Unnecessary Article (an)

dhurlburtusa:

The problem is extraneous commas and an unnecessary article (an) in the documentation.


esasse:
@bundlerbot r+

esasse:
Thanks @dhurlburtusa !

PR #500, opened 2020-01-23
Fix canonical linking

deivid-rodriguez:

The problem was that Google is showing bundler docs for old versions in search results. An example of this problem is pointed out by @indirect in the 3rd bullet point of https://github.com/rubygems/bundler/pull/7591#discussion_r369845718.

My diagnosis was that we already have a solution for this, namely, consider the latest version of each page as the canonical version.

The problem, I think, is that the regexp for adding the canonical link was incorrect for the kind of pages referenced by @indirect.

My fix is to use a proper regexp that includes all the pages where we want to do this. While fixing it, I also noticed other pages where we also want to do this that were using a different layout.

I chose this fix because it should improve our docs visibility in search results.


indirect:
Oh wow! I didn't even realize we could fix this so easily, thank you! 🎉 

@bundlerbot merge

PR #501, opened 2020-01-24
Expand FAQs about Gemfile and Gemfile.lock usage in libraries

deivid-rodriguez:

The problem was that the information about the reasons to commit your `Gemfile.lock` could be deepened and clarified.

My diagnosis was that we should expand on the reasons. Also, I noticed that we're not mentioning the roles on `Gemfile` and `Gemfile.lock` files inside gems, which is something not very well known.

My fix is to try to clarify the reasons for committing `Gemfile.lock` by default, and also to add a specific question about the roles of `Gemfile` and `Gemfile.lock` inside gems.

### DIFF
--- a/source/v2.0/guides/faq.html.haml
+++ b/source/v2.0/guides/faq.html.haml
@@ -175,18 +175,104 @@ title: Frequently Asked Questions
           $ bundle install --full-index
 
         %h3
-          Committing lockfiles in libraries
+          Using Gemfile and Gemfile.lock in libraries
+
+        %p
+          <strong>Q</strong>: How are Gemfile{,.lock} files used when inside a library?
+
+        %p
+          <strong>A</strong>: This is something not very well known. When end
+          users install your library, the `Gemfile` and `Gemfile.lock` files
+          inside your repo are completely ignored, no matter whether you are
+          shipping them with your gem or not.
+
+          They are only useful for installing your gemspec dependencies in a
+          development environment, and also to include additional dependencies
+          only for development / testing.
+
+          The following is a possible `Gemfile` inside a gem:
+
+          :code
+            source "https://rubygems.org"
+
+            gemspec
+
+            gem "rspec", "~> 3.9"
+            gem "rubocop", "0.79.0"
+
+          The `gemspec` DSL makes sure `bundler` considers the dependencies in
+          your `.gemspec` file for resolution, but again: any dependencies
+          outside of your gemspec file are only used for development and never
+          affect final users of your gem.
+
+          The same thing happens with the development dependencies defined in your
+          gemspec through `Gem::Specification#add_development_dependency`.
 
         %p
           <strong>Q</strong>: Should I commit my `Gemfile.lock` when writing a gem?
 
         %p
-          <strong>A</strong>: Yes. When Bundler first shipped, the
-          `Gemfile.lock` was gitignored inside gems.  Over time, however, it
-          became clear that this practice forces the pain of broken dependencies
-          onto new contributors, while leaving existing contributors potentially
-          unaware of the problem. Since `bundle install` is usually the first
-          step towards a contribution, the pain of broken dependencies would
-          discourage new contributors from contributing. As a result, we have
-          revised our guidance for gem authors to now recommend checking in the
-          lock for gems.
+          <strong>A</strong>: Our advice: yes. The presence of a `Gemfile.lock`
+          in a gem's repository ensures that a fresh checkout of the repository
+          uses the exact same set of dependencies every time. We believe this
+          makes repositories more friendly towards new and existing
+          contributors.
+
+          Ideally, contributors should be able to work over a clean slate by
+          default. So at any given time, cloning the repo, running `bundle
+          install` and then running your gem's test suite should succeed.
+          Otherwise it can get discouraging, specially for beginners.
+
+        %p
+          <strong>Q</strong>: What happens if I don't commit my lockfile to source control when writing a gem?
+
+        %p
+          <strong>A</strong>: If you don't commit your lockfile to source
+          control, your development environment loses this determinism because:
+
+          %ul
+            %li
+              The <code>bundle install</code> step could start failing if you
+              pull new code with changes in the Gemfile. Your current lockfile
+              might get out of date, and bundler would find conflicting
+              requirements in both files and fail.
+
+            %li
+              Even if <code>bundle install</code> succeeds, your tests could
+              start failing because of changes on third party development or
+              runtime dependencies not under your control.
+
+          Some people argue that these are a slight benefit, since they may
+          cause you to find about broken dependencies earlier. The bundler team,
+          however, considers that this is not an issue in most situations
+          because:
+
+          %ul
+            %li
+              Changes in external dependencies _should not_ break your gem. As a
+              gem maintainer, you should find out about your dependencies' semver
+              policy and restrict your requirements accordingly in your gemspec
+              file.
+
+            %li
+              In real life, sometimes changes in external dependencies _do_
+              (unintentionally) cause breakage. If this happens on a development
+              dependency (for example, your testing gem makes a bad release
+              preventing your tests from running), this is not a real issue with
+              your gem, but still harms contributors and the stability of your
+              CI. By committing the lockfile, maintainers can upgrade the
+              development dependency and fix the issue at their own pace, while
+              CI and contributors don't get affected. If it's a runtime
+              dependency causing the breakage, our experience tells us that it's
+              most likely going to be your end users who will find out about the
+              issue and report it anyways, no matter whether you commit your
+              lockfile or not.
+
+            %li
+              There's other good practices that can keep the predictability of
+              committing your lockfile to source control, but still allow you to
+              proactively run your tests against the latest releases of your
+              dependencies and finding out about problems. For example,
+              [dependabot] can create PRs for you in real time as new versions
+              of your dependencies are released, so that as a gem mantainer you
+              can be the first one to find about any issues :)
--- a/source/v2.0/guides/faq.html.haml
+++ b/source/v2.0/guides/faq.html.haml
@@ -222,13 +222,12 @@ title: Frequently Asked Questions
 
           Instead of forcing every fresh checkout (and possible new
           contributor) to encounter broken builds, the Bundler team recommends
-          creating a CI configuration that deletes your `Gemfile.lock`, runs
-          `bundle install`, and then runs your tests. That way you, and others
-          monitoring the build status, will be the first to know about failures
-          from dependency changes.
+          either using a tool like <a href="https://dependabot.com">Dependabot
+          </a> to automatically create a PR and run the test suite any time
+          your dependencies release new versions.
 
-          To make updating dependencies less onorous, use tools like <a
-          href="https://dependabot.com">Dependabot</a> to automatically run
-          tests against new versions of your dependencies, and to easily (or
-          even automatically!) update your Gemfile.lock to the latest
-          verisons of your dependencies.
+          If you don't want to use a dependency monitoring bot, we suggest
+          creating an additional daily CI build that deletes the Gemfile.lock
+          before running `bundle install`. That way you, and others monitoring
+          your CI status, will be the first to know about any failures from
+          dependency changes.
--- a/source/v2.0/guides/faq.html.haml
+++ b/source/v2.0/guides/faq.html.haml
@@ -190,10 +190,9 @@ title: Frequently Asked Questions
           on your gem. The `Gemfile` also provides an easy way to track and
           install development-only or test-only gems.
 
-          Read more about `gemspec` in Gemfiles from the <a
-          href="../rubygems.html">RubyGem reference page</a> and the <a
-          href="./creating_gem.html">How to create a gem with Bundler</a>
-          guide.
+          Read about Gemfiles in gems from the <a href="../rubygems.html">
+          Bundler in gems</a> page and the <a href="./creating_gem.html">
+          How to create a gem with Bundler</a> guide.
 
         %p
           <strong>Q</strong>: Should I commit my `Gemfile.lock` when writing a gem?
--- a/source/v2.0/guides/faq.html.haml
+++ b/source/v2.0/guides/faq.html.haml
@@ -175,104 +175,61 @@ title: Frequently Asked Questions
           $ bundle install --full-index
 
         %h3
-          Using Gemfile and Gemfile.lock in libraries
+          Using Gemfiles inside gems
 
         %p
-          <strong>Q</strong>: How are Gemfile{,.lock} files used when inside a library?
+          <strong>Q</strong>: What happens if I put a `Gemfile` in my gem?
 
         %p
-          <strong>A</strong>: This is something not very well known. When end
-          users install your library, the `Gemfile` and `Gemfile.lock` files
-          inside your repo are completely ignored, no matter whether you are
-          shipping them with your gem or not.
+          <strong>A</strong>: When someone installs your gem, the `Gemfile` and
+          `Gemfile.lock` files are completely ignored, even if you include them
+          inside the `.gem` file you upload to rubygems.org.
 
-          They are only useful for installing your gemspec dependencies in a
-          development environment, and also to include additional dependencies
-          only for development / testing.
+          The `Gemfile` inside your gem is only to make it easy for developers
+          (like you) to install the dependencies needed to do development work
+          on your gem. The `Gemfile` also provides an easy way to track and
+          install development-only or test-only gems.
 
-          The following is a possible `Gemfile` inside a gem:
-
-          :code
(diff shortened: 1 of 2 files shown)


source/v2.0/guides/faq.html.haml
```diff
@@ -175,18 +175,105 @@ title: Frequently Asked Questions
...
+            gem "rspec", "~> 3.9"
+            gem "rubocop", "0.79.0"
+          ```
+
+          The `gemspec` DSL makes sure `bundler` considers the dependencies in
+          your `.gemspec` file for resolution, but again: any dependencies
+          outside of your gemspec file are only used for development and never
+          affect final users of your gem.
+
+          Same thing happens with the development dependencies defined in your
```
ianfixes: Nit: _The_ same thing happens.

With respect to `add_development_dependency`, this seems like a can of worms.  Should this be mentioned in the docs for `add_development_dependency`? Like, is that function deprecated?  What is the difference under the hood between putting a Gem's development dependencies in the `.gemspec` vs the `Gemfile`?

deivid-rodriguez: We considered deprecating it but we decided it was not worth the trouble at this point. So what we're doing for the time being is spreading the word that using `Gemfile` is better for this matter.

They are essentially the same thing but using a `Gemfile` is more flexible, since you can point to `git` sources, for example.

ianfixes:
My motivation for opening #487 was to address how these opinions might apply differently to the situation of private gems and/or a closed development team. 

I'm thinking that you should split the answer to this question:

> Should I commit my `Gemfile.lock` when writing a gem?

To answer a separately-asked question like:

> What should I expect if I _don't_ commit my `Gemfile.lock`?

I'll try to write a quick summary of what that would look like, give me a few minutes.

ianfixes:
### Should I commit my `Gemfile.lock` when writing a gem?

Our advice: yes. The presence of a `Gemfile.lock` blah blah blah

Ideally, contributors should be able to work over a clean slate by blah blah blah

Otherwise it can get discouraging, specially for beginners. 


### What happens if I don't commit my lockfile to source control when writing a gem?

If you don't commit your lockfile to source control,

* The `bundle install` step could start failing if you `git pull` blah blah blah
* Even if `bundle install` succeeds, your tests could start failing blah blah blah

Some people argue that these are a slight benefit, since they may cause you to find about broken dependencies earlier.

### What is the bundler team's reasoning on the issue of committing lockfiles when writing a gem?

* Changes in external dependencies _should not_ break your gem. blah blah blah
* In real life, sometimes changes in external dependencies _do_ blah blah blah
* There's other good practices that can keep the predictability of blah blah blah

deivid-rodriguez:
The thing is... Even if working on a private gem, or with a closed development team, I would still commit the `Gemfile.lock` file. Even if everybody is ready to fix issues in broken dependencies, I think there's other better ways to deal with that.

Also, in order to discover these problems locally, developers would need to constantly delete their `Gemfile.lock` and re-resolve, since I assume all developers in the closed team will already have a `Gemfile.lock` copy.

I feel there are many specific situations here so I think it's enough to say that this is our opinion, but it's cool to disagree (and really simple: add it gitignore, and `git rm`).

ianfixes:
Our comments crossed.  I didn't change what you wrote, just broke it into sections and I think it addresses my concerns perfectly that way.

deivid-rodriguez:
Haha, yeah, comments crossed :)

Ok, I partially incorporated your feedback, but splitting it only in two questions, not three, since I feel the first and third questions are essentially the same.

source/v2.0/guides/faq.html.haml
```diff
@@ -175,18 +175,108 @@ title: Frequently Asked Questions
...
+          <strong>A</strong>: This is something not very well known. When end
+          users install your library, the `Gemfile` and `Gemfile.lock` files
+          inside your repo are completely ignored, no matter whether you are
+          shipping them with your gem or not.
+
+          They are only useful for installing your gemspec dependencies in a
+          development environment, and also to include additional dependencies
+          only for development / testing.
+
+          Thw following is a typical `Gemfile` inside a gem:
```
AlexWayfer: A typo, I guess.

source/v2.0/guides/faq.html.haml
```diff
@@ -175,18 +175,108 @@ title: Frequently Asked Questions
...
+          The `gemspec` DSL makes sure `bundler` considers the dependencies in
+          your `.gemspec` file for resolution, but again: any dependencies
+          outside of your gemspec file are only used for development and never
+          affect final users of your gem.
+
+          The same thing happens with the development dependencies defined in your
+          gemspec through `Gem::Specification#add_development_dependency`. As a
+          matter of fact, the bundler team recommends to not use
+          `add_development_dependency` and instead move development dependencies
+          to your Gemfile.
```
AlexWayfer: Just unreasoned recommendation?

source/v2.0/guides/faq.html.haml
```diff
@@ -175,18 +175,108 @@ title: Frequently Asked Questions
...
+          makes repositories more friendly towards new and existing
+          contributors.
+
+          Ideally, contributors should be able to work over a clean slate by
+          default. So at any given time, cloning the repo, running `bundle
+          install` and then running your gem's test suite should succeed.
+          Otherwise it can get discouraging, specially for beginners.
+
+        %p
+          <strong>Q</strong>: What happens if I don't commit my lockfile to source control when writing a gem?
```
AlexWayfer: You wrapped answers texts, but not (this) question. Maybe a mistake, I'm not sure.

source/v2.0/guides/faq.html.haml
```diff
@@ -175,18 +175,108 @@ title: Frequently Asked Questions
...
+
+          %ul
+            %li
+              Changes in external dependencies _should not_ break your gem. As a
+              gem maintainer, you should find out about your dependencies' semver
+              policy and restrict your requirements accordingly in your gemspec
+              file. However, restricting them too much makes your library less
+              flexible when being used together with other libraries with the
+              same dependencies, increasing the chance of resolution conflicts
+              in end applications. Normally using the `~>` operator is a good
```
AlexWayfer: > you should find out … and restrict requirements

Sounds right.

> increasing the chance of resolution conflicts in end applications

And therefore let's lock dependencies not so strict without dependencies conflicts but with ability to break applications! Sounds… weird.

source/v2.0/guides/faq.html.haml
```diff
@@ -175,18 +175,108 @@ title: Frequently Asked Questions
...
+              compromise and will prevent external dependencies from breaking
+              your gem in the majority of the cases.
+
+            %li
+              In real life, sometimes changes in external dependencies _do_
+              (unintentionally) break your gem. If this happens on a development
+              dependency, this is completely inocous for your end users, but
```
AlexWayfer: `innocuous`?

source/v2.0/guides/faq.html.haml
```diff
@@ -175,18 +175,108 @@ title: Frequently Asked Questions
...
+              still harms contributors and the stability of your CI, so avoiding
+              that is always good. If it happens on a runtime dependency, our
```
AlexWayfer: > so avoiding that is always good

Why? It's broken already, why closing eyes is good when the problem is still here?

source/v2.0/guides/faq.html.haml
deivid-rodriguez: Yeah.

ianfixes:
> two questions, not three

I agree that the first and third questions are similar, but "What happens if I don't commit my lockfile?" is pretty objective question.  You give it an objective answer.

But then you go on to say "here's why our team doesn't care what happens when you don't commit  your lockfile"... which is subjective.  So my thinking was to split the answer into its objective and subjective parts by having an objective and a subjective question to introduce them.

At the moment I should say that I am satisfied either way with the overall content.

source/v2.0/guides/faq.html.haml
deivid-rodriguez: Yeah, I dropped it. Will consider recommending against it with more explanations another time.

source/v2.0/guides/faq.html.haml
deivid-rodriguez: Fixed, I think.

source/v2.0/guides/faq.html.haml
deivid-rodriguez: Not sure what you mean, are you proposing that we recommend locking exact versions in gemspecs? :grimacing:

source/v2.0/guides/faq.html.haml
AlexWayfer: OK. Also, I'm not sure about "typical `Gemfile`". If I'm using `add_development_dependency` — my typical `Gemfile` is `source` + `gemspec` and that's all.

source/v2.0/guides/faq.html.haml
AlexWayfer: OK, let's clear what is "restricting them too much"?

source/v2.0/guides/faq.html.haml
deivid-rodriguez: Thanks! I reword the whole thing and I'm no longer using that word, but I learnt the spelling :)

source/v2.0/guides/faq.html.haml
deivid-rodriguez: Thanks for the feedback, it means I didn't communicate what I wanted properly. I'll reword it.

source/v2.0/guides/faq.html.haml
deivid-rodriguez: Ok, I'll say "a possible" `Gemfile`.

source/v2.0/guides/faq.html.haml
deivid-rodriguez: I dropped the part about restricting them too much, maybe I'll create some more docs about "good practices for gemspec requirements" other time.

source/v2.0/guides/faq.html.haml
deivid-rodriguez: Ended up dropping that part, since it probably requires more elaboration.

source/v2.0/guides/faq.html.haml
AlexWayfer: Doesn't look so:

image

source/v2.0/guides/faq.html.haml
```diff
@@ -175,18 +175,104 @@ title: Frequently Asked Questions
...
+
+          The following is a typical `Gemfile` inside a gem:
+
+          :code
+            source "https://rubygems.org"
+
+            gemspec
+
+            gem "rspec", "~> 3.9"
+            gem "rubocop", "0.79.0"
```
AlexWayfer: To not forget, from https://github.com/rubygems/bundler-site/pull/501#discussion_r370651889

> Ok, I'll say "a possible" Gemfile.

deivid-rodriguez:
@ianfixes I see, thanks! I'll leave the PR rest in there for a few days to see if I can come up with a better structure or question wording to improve that. I'll leave it as it is otherwise.

source/v2.0/guides/faq.html.haml
deivid-rodriguez: Thanks, I had indeed forgotten!

AlexWayfer:
> If it's a runtime dependency causing the breakage, our experience tells us that it's most likely going to be your end users who will find out about the issue and report it anyways, no matter whether you commit your lockfile or not.

Good part. Except cases when maintenance is active and you (gem maintainer) catch breakages before releases. But for slowly-maintained gems it doesn't matter, I guess.

source/v2.0/guides/faq.html.haml
```diff
@@ -175,104 +175,61 @@ title: Frequently Asked Questions
...
+          Gemfile.lock is that new checkouts (including CI) will immediately
+          have failing tests if one of your dependencies changes in a breaking
+          way.
+
+          Instead of forcing every fresh checkout (and possible new
+          contributor) to encounter broken builds, the Bundler team recomments
+          creating a CI configuration that deletes your `Gemfile.lock`, runs
+          `bundle install`, and then runs your tests. That way you, and others
+          monitoring the build status, will be the first to know about failures
+          from dependency changes.
```
deivid-rodriguez: I'm personally fine with this, as long as it doesn't make PRs get red CI status due to dependency changes, i.e., as long as this is configured as a scheduled run, or something that notifies maintainers, but doesn't affect regular PRs.

In my personal view, a PR should have a red status if and only if the specific changes in the PR make the CI fail.

source/v2.0/guides/faq.html.haml
```diff
@@ -175,104 +175,61 @@ title: Frequently Asked Questions
...
-              [dependabot] can create PRs for you in real time as new versions
-              of your dependencies are released, so that as a gem mantainer you
-              can be the first one to find about any issues :)
+          <strong>A</strong>: The main advantage of not checking in your
+          Gemfile.lock is that new checkouts (including CI) will immediately
+          have failing tests if one of your dependencies changes in a breaking
+          way.
+
+          Instead of forcing every fresh checkout (and possible new
+          contributor) to encounter broken builds, the Bundler team recomments
```
deivid-rodriguez:
```suggestion
          contributor) to encounter broken builds, the Bundler team recommends
```

source/v2.0/guides/faq.html.haml
ianfixes: > a PR should have a red status if and only if the specific changes in the PR make the CI fail

I'll agree to disagree.  While it may not seem "fair" for a contributor to see their build fail due to external factors, it would be dishonest for the CI to misrepresent the outcome of tests, or to avoid reporting on certain tests that fail.

That said, you should expect your CI system to be able to communicate whether a failure is due to internal or external factors.  E.g. using Travis to run the test of dependencies as a separate parallel build. 

Note that it is the system of bundling Ruby gems itself that produces these kinds of failures -- by design; the CI is just the messenger.

source/v2.0/guides/faq.html.haml
```diff
@@ -175,18 +175,61 @@ title: Frequently Asked Questions
...
+        %p
+          <strong>A</strong>: When someone installs your gem, the `Gemfile` and
+          `Gemfile.lock` files are completely ignored, even if you include them
+          inside the `.gem` file you upload to rubygems.org.
+
+          The `Gemfile` inside your gem is only to make it easy for developers
+          (like you) to install the dependencies needed to do development work
+          on your gem. The `Gemfile` also provides an easy way to track and
+          install development-only or test-only gems.
+
```
ianfixes: The `Gemfile` also provides an easy way to track and install development-only or test-only gems.  For example, `rspec` and `rubocop`:

          :code
            source "https://rubygems.org"

            gemspec

            gem "rspec", "~> 3.9"
            gem "rubocop", "0.79.0"

source/v2.0/guides/faq.html.haml
deivid-rodriguez: Yes, I'll agree to disagree too :) In my opinion, red PRs where "CI failures seem unrelated" are just annoying and in the great majority of the cases do not lead to maintainers finding out about a problem before anyone else. I cannot recall a single time when a failure of this kind actually made me aware of an actually problem with my library.

ianfixes: > red PRs where "CI failures seem unrelated" are just annoying 

Yeah, don't get me started.  This is a limitation of our industry, where we've decided to roll up the entire complexity of whether to accept a patch into a single boolean, when (as is evident here) there are really many independent facets of the codebase that a given patch can affect.  

I think we agree on the lack of nuance conveyed by a build status but not on whether to "round up" or "round down" 😄

deivid-rodriguez: I'm sure everyone agrees that CI is just a guideline but not a definitive indicator of the acceptance of a PR.

Anyways, I'm not sure whether I'm rounding down or up here, since my personal best practice is to actually get a PR created in real time right after any dependency is released :sweat_smile:. Might be overkill, but if it's about finding broken dependencies it seems quite optimal.

source/v2.0/guides/faq.html.haml
indirect: I don't think an entire example gemfile is a good fit for a FAQ answer, but I think it does make sense to give an example and explain it in in the page about using Bundler while developing gems. I've added it over there. 👍🏻

source/v2.0/guides/faq.html.haml
indirect: Updated to mention Dependabot etc as a first choice and a _secondary_ CI build as second choice. 😅

indirect:
@deivid-rodriguez feel free to make any other changes you think this could use, it's r+ from me now 👍

deivid-rodriguez:
Nice, I'm super happy with how this reads now, thanks @indirect  :purple_heart:

Also thanks @ianfixes for all the good feedback and for getting this started!

@bundlerbot merge

PR #502, opened 2020-02-13
Normalize case

deivid-rodriguez:

The problem was that while working on canonical links, I was grepping for "Learn More" links and missing a few expected results.

My diagnosis was that "Learn More" links have inconsistent casing.

My fix is to make it consistent.

I chose this fix because it makes the site's look more uniform and not subject to grepping surprises. I'm also happy with the "Inverse fix" of changing everything to "Learn more".


deivid-rodriguez:
A bit of a dummy PR, but I'll merge it :)

deivid-rodriguez:
@bundlerbot merge

Issue #505, opened 2020-03-22
Known bundler plugins


bronzdoc:
"This would be a page on the Bundler website, with a rake task we can run to query rubygems and update the list of plugins and link to their gem pages."

See: https://github.com/rubygems/rubygems/issues/3376

olleolleolle:
I'll try to do something about this tonight.

tnir:
Hi @olleolleolle, do you have any plan to complete this? You just stated the problem in #505 as below but I am not able to find the page on the Web site:

> See #505 - there is no Known Plugins page in the site.

olleolleolle:
@tnir oh,  thanks for asking, I won't be doing anything about this, now.

tnir:
Okay, @olleolleolle. Remove you from assignees and push it to https://github.com/rubygems/bundler-site/contribute.

PR #510, opened 2020-04-08
Add data/known_plugins.yml and its Rake task

olleolleolle:

See #505 - there is no **Known Plugins** page in the site. Users have no way of discovering the community's Plugins for Bundler.

#505 calls for a regeneratable YAML data-set for Middleman, kept up-to-date with a Rake task.

The generated YAML file could be checked in to git. (The Issue calls for this, too.)

❓ **Wondering:**
Most of the gems named with a prefix `bundler-` are not using the `plugins.rb` way of extending Bundler, but the "add a CLI command" way.

This PR currently adds data for `plugins.rb`-using gems, only. I'm wondering if you think I should add fewer skips.

This PR adds a YAML file from which it's easy to add a page. Compare to the Contributors page. Note: this PR does not include adding a page, only fetching the data for one.

My fix was "use some curl and grep" from the existing Issues, wrap that in a Rake task, and add a lot of ignored gem names.

#### TODO

- [x] Go through each of the elements in the generated YAML and check for a `plugins.rb` being there

I chose this fix because I noted that the data/ directory was already there, and that we are tending this data ourselves. We can tend another set of data, too. Esp. with explicit skips.

### DIFF
--- a/data/known_plugins.yml
+++ b/data/known_plugins.yml
@@ -1,0 +1,21 @@
+---
+- :name: bundler-changelogs
+  :uri: https://github.com/jdar/bundler-changelogs
+- :name: bundler-console
+  :uri: https://github.com/kddeisz/bundler-console
+- :name: bundler-dependency_graph
+  :uri: https://github.com/kerrizor/bundler-dependency_graph
+- :name: bundler-ecology
+  :uri: https://github.com/eco-rb/bundler-ecology
+- :name: bundler-explain
+  :uri: https://github.com/jhawthorn/bundler-explain
+- :name: bundler-inject
+  :uri: https://github.com/ManageIQ/bundler-inject
+- :name: bundler-licensed
+  :uri: https://github.com/sergey-alekseev/bundler-licensed
+- :name: bundler-private_install
+  :uri: https://github.com/fphilipe/bundler-private_install
+- :name: bundler-source-aws-s3
+  :uri: https://github.com/eki/bundler-source-aws-s3
+- :name: bundler-symlink
+  :uri: https://github.com/petekinnecom/bundler-symlink
--- a/lib/tasks/regenerate_known_plugins_yml.rake
+++ b/lib/tasks/regenerate_known_plugins_yml.rake
@@ -1,0 +1,91 @@
+desc "Recreate data/known_plugins.yml from RubyGems.org data. " \
+  "Look for gems with bundler- prefix names, and get their information, " \
+  "skipping invalid gems. Run with RUBYOPT=-W1 enabled to see diagnostic " \
+  "output about data not found. Uses curl, grep."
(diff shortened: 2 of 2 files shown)


data/known_plugins.yml
```diff
@@ -0,0 +1,45 @@
+---
+- :name: bundler-advise
+  :uri: https://github.com/chrismo/bundler-advise
+- :name: bundler-audit
+  :uri: https://github.com/rubysec/bundler-audit#readme
+- :name: bundler-audited_update
+  :uri: http://rubygems.org/gems/bundler-audited_update
```
bronzdoc, line 7: I think, this is not a plugin either

data/known_plugins.yml
```diff
@@ -0,0 +1,45 @@
...
+- :name: bundler-audit-ng
+  :uri: https://github.com/pxlpnk/bundler-audit-ng
```
bronzdoc, line 9: This is just a fork of bundler audit 🤔

lib/tasks/regenerate_known_plugins_yml.rake
```diff
@@ -0,0 +1,79 @@
...
+    bundler-changelog empty
+    bundler-fastupdate
+    bundler-fu
+    bundler-gem-hg
+    bundler-geminabox
+    bundler-gem_version_tasks
+    bundler-github functionality-exists-in-bundler-now
+    bundler-bouncer
+    bundler-interactive source-does-not-exist yanked-all-but-last
+    bundler-maglev-
```
bronzdoc, line 24: That las `-` is intentional?

olleolleolle: Yes that is the name of it - and it seems fishy - looks like a fork of bundler source code. No explanations given.

data/known_plugins.yml
```diff
@@ -0,0 +1,45 @@
+---
+- :name: bundler-advise
+  :uri: https://github.com/chrismo/bundler-advise
+- :name: bundler-audit
```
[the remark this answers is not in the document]
olleolleolle: Thanks, dropped. (There are probably more.)

data/known_plugins.yml
olleolleolle: Thanks, dropped.

data/known_plugins.yml
olleolleolle: Thanks, dropped.

olleolleolle:
@bronzdoc There, second pass done, and now only plugins.rb-having gems are listed in the data file.

data/known_plugins.yml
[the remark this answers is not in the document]
olleolleolle: :warning: I should drop this.

bronzdoc:
Thanks @olleolleolle!

PR #509, opened 2020-04-08
Update to Ruby 2.6.6

esasse:
It's always good to use a newer version. 😀

PR #512, opened 2020-04-09
chore: Update Gemfile.lock to reflect Ruby 2.6.6p146

olleolleolle:

The problem was as a developer, I'd get a change in the Gemfile after `bundle install`.

My diagnosis was that this could be a problem for the bundle install `--deployment` flag, which forbids changes to the lockfile on installing.

A recent update to the Ruby version used made this change necessary. See #509

My fix was to check in the current version of the lockfile.

I chose this fix because I couldn't come up with any alternative.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -231,7 +231,7 @@ DEPENDENCIES
   travis
 
 RUBY VERSION
-   ruby 2.6.3p62
+   ruby 2.6.6p146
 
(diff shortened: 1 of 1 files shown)


olleolleolle:
I can see now that the Travis CI task is using its preinstalled `Bundler version 1.17.2`, so I will edit the Gemfile.lock I created here.

PR #513, opened 2020-04-09
Git Sources shorthands: Deprecation note

### LINKED ISSUE #503: Installing gems from gists is underdocumented.
https://github.com/rubygems/bundler-site/blob/b86f2ccd3150009bd9bb25d961a462a880631116/source/v2.0/guides/git.html.haml#L111-L121

Is this even possible, considering gists don't support directories? Was it ever?

olleolleolle:

The problem was that the deprecated status of the built-in shorthands was not documented.

Fixes #503

My diagnosis was that @duckinator was right about the way to deal with the shorthands in the documentation.

My fix was to **document the deprecation**  (v2.0/ directory) with a *NB* marker.

**Update**: Added "what to do" as a suggestion.

I chose this fix because I didn't know about any special tricks to show "All This Section Is Deprecated".

### DIFF
--- a/source/v2.0/guides/git.html.haml
+++ b/source/v2.0/guides/git.html.haml
@@ -86,6 +86,8 @@ title: How to install gems from git repositories
         gem 'rugged', git: 'git://github.com/libgit2/rugged.git', submodules: true
     .bullet
       .description
+        (<b>NB</b>: These built-in shorthand git sources are deprecated and will be
+        removed in Bundler 3.)
         If you are getting your gems from a public GitHub repository,
         you can use the shorthand
       :code
--- a/source/v2.0/guides/git.html.haml
+++ b/source/v2.0/guides/git.html.haml
@@ -87,7 +87,8 @@ title: How to install gems from git repositories
     .bullet
       .description
         (<b>NB</b>: These built-in shorthand git sources are deprecated and will be
-        removed in Bundler 3.)
+        removed in Bundler 3. Instead of these, we encourage you to use <em>Custom
+        git sources</em> in your Gemfile.)
         If you are getting your gems from a public GitHub repository,
         you can use the shorthand
       :code


source/v2.0/guides/git.html.haml
```diff
@@ -86,6 +86,8 @@ title: How to install gems from git repositories
         gem 'rugged', git: 'git://github.com/libgit2/rugged.git', submodules: true
     .bullet
       .description
+        (<b>NB</b>: These built-in shorthand git sources are deprecated and will be
+        removed in Bundler 3.)
```
olleolleolle, line 90: Perhaps this sentences reads easier with an exhortation to do a new thing: "We encourage you to use custom shorthand git sources in your Gemfile."

simi: I agree, it will be nice and more friendly to also explain what is "replacement" for this.

olleolleolle: About now?

simi:
From another point of view, if we would like to not promote this, wouldn't be less confusing to just remove this example?

olleolleolle:
I think it's important that users can look in the docs, learn what it is they see in some project's Gemfile. Even if that feature will be going away in the future.

That said, perhaps this is a case for switching the order around. "Custom git sources" could be placed first, and then, as an afterthought, the Shorthand built-in git sources.

simi:
sounds good to me

deivid-rodriguez:
I believe we need to fix something here in bundler, because using these source doesn't currently print any deprecation warnings in bundler 2, but they are broken in bundler 3.

So we need to either fix bundler to print deprecations for them, or to get them back to working and not deprecate them at all.

deivid-rodriguez:
Ok, I'm digging through history now and we decided to delay the deprecation of these to bundler 3: https://github.com/rubygems/bundler/pull/7000.

The rationale if I recall correctly was that we were also in the middle of switching these sources to use `https` instead of `git` under the hood by default, so deprecating everything at the same time was a bit of a mess.

Well, and honestly, I kind of like these shortcuts, I'm not sure about the reason for deprecating them but I assume it was a good reason, so I didn't dig deeper.

Anyways, there's a bug in bundler since in bundler 3, these sources are broken like this:

```
 $ bundle

[!] There was an error parsing `Gemfile`: You passed :github as an option for gem 'rails', but it is invalid. Valid options are: group, groups, git, path, glob, name, branch, ref, tag, require, submodules, platform, platforms, type, source, install_if, gemfile. You may be able to resolve this by upgrading Bundler to the newest version.. Bundler cannot continue.

 #  from /home/deivid/Code/playground/bundler/shortcuts/Gemfile:5
 #  -------------------------------------------
 #  
 >  gem "rails", github: "rails/rails"
 #  -------------------------------------------
```
where the expectation is to only warn.

Regarding documenting this at this stage, I don't have a strong opinion.

olleolleolle:
Thanks for all the review effort! I will close this and not change anything at this point, we can return to it when Bundler 3 is coming around, and then we'll have all these words & thoughts to look at, if we ever need them. Ciao!

PR #515, opened 2020-04-21
Update bundle config syntax to non-deprecated version

jm3:
[DEPRECATED] Using the `config` command without a subcommand [list, get, set, unset] is deprecated and will be removed in the future. Use `bundle config set local.gem_name /foo/bar/baz.git` instead.

The problem was that docs are out of sync with the current version of bundler.

Running the commands as given in the README gives a deprecation error.

Update the docs

It is good.

### DIFF
--- a/source/v2.0/guides/git.html.haml
+++ b/source/v2.0/guides/git.html.haml
@@ -148,12 +148,12 @@ title: How to install gems from git repositories
         instead of using the remote version. This can be achieved by setting
         up a local override:
       :code
-        $ bundle config local.GEM_NAME /path/to/local/git/repository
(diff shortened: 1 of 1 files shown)


olleolleolle:
https://github.com/rubygems/bundler-site/pull/515/checks?check_run_id=603975140#step:6:1374 the CI that ran middleman managed to dump its core, wow. Well, unrelated, to say the least.

olleolleolle:
Attempting a re-run again.

olleolleolle:
(There'd been a commit to master which activated builds on PRs in some way, perhaps that had a glitch. I merged this content change.)

PR #517, opened 2020-04-28
Adjust example Gemfile to best practices

vog:

I saw the example `Gemfile` on the main website, which uses a global `source` command, clicked on "Learn more", just to learn that global `source` commands are strongly discouraged for good reasons.

The main website violated best practices, so it probably hasn't received much attention over the last weeks or months.

Adjust the example to the best practices.

The alternative would be to change the recommendation on the Gemfile description site, but the reasons provided for discouraging global `source` commands are very convincing.

### DIFF
--- a/source/v2.1/gemfile.html.haml
+++ b/source/v2.1/gemfile.html.haml
@@ -16,11 +16,20 @@
         Gemfiles require at least one gem source, in the form of the URL for a RubyGems server. Generate a Gemfile with the default rubygems.org source by running <code>bundle init</code>. If you can, use <code>https</code> so your connection to the rubygems.org server will be verified with SSL.
       :code
         # lang: ruby
-        source 'https://rubygems.org' do
-          # Gems here
-        end
+        source 'https://rubygems.org'
+
+        # Gems here
       .notes
-        Global source lines are a security risk and should not be used as they can lead to gems being installed from unintended sources.
+        If you want to pick up gems from a different source, make sure to wrap the extra sources in a block. Multiple global source lines are a security risk and should not be used as they can lead to gems being installed from unintended sources.
+      :code
+        # lang: ruby
+        source 'https://rubygems.org'
+
+        # Gems here
+
+        source 'https://gems.example.com' do
+          # Gems from the alternative source here
+        end
 
     .bullet
       .description


vog:
I don't understand why this simple change made the build fail. The build error message is also very strange:

```
You may have encountered a bug in the Ruby interpreter or extension libraries.
Bug reports are welcome.
```

simi:
:wave: @vog, no worries. Build fail is not introduced by your change.

olleolleolle:
Perhaps the messaging about global `source` statements could be made more precise, to read something such as "Avoid global `source` statements that aren't referring to `https://rubygems.org`"?

And, perhaps some wording I currently don't have, which would explain why life would be harder if I added my second global `source "https://olleolleolle-gems.com:9292"` (as opposed to putting its `gem` statements in a `source` block).

By this I suggest that we keep the Gemfile example the way it was before this PR, but that the other text could change.

Does that rhyme with other Gemfiles you have seen used in the  world?

vog:
Assuming that a global `source` of `https://rubygems.org` is harmless, I still don't se why we would want to introduce that kind of special rule. Are the saving (one `do` and one `end`) really worth the trouble? The main example then states an exception rather than a rule, and almost invites to add further sources via global `source` as well.

I would find it much easier to understand, especially for people new to `bundler` (which should be the target group of the main website), to only have `source ... do`, which conveys clear, safe and straight forward way to extent to different sources.

simi:
@vog I think it is not about considering `https://rubygems.org` harmless. Having one source on the top of Gemfile and do not use any other is harmless, since all gems share the same source. The problem occurs when you start to mixing sources since it can easily result into gem being downloaded from different source than expected (which is security risk).

```ruby
# Gemfile
source 'https://rubygems.org'
source 'https://olleolleolle-gems.com:9292'

gem 'my-gem' # which source would be used if gem is present in both sources?
```

Anyway I understand your confusion and we should definitely improve at least texts at website. If we would like to move to block source as suggested one by default, we should also do this change to major Gemfile templates (like `bundle init`, `rails new', ...).

indirect:
A single global source is always fine, so I think the initial example is OK. The problem is when you have more than one source in a single Gemfile, it is important to know which source each gem comes from.

I think the official recommendation is for every source _besides_ your main source to have a `do/end` block?

deivid-rodriguez:
Hello @vog!

We haven't heard back from you, so I'll update this PR with the "inverse change". In this case, it's the recommendation after you click on "Learn more" that it's wrong as pointed out by other maintainers, not the example Gemfile.

Thanks for bringing this to our attention!

olleolleolle (approved):
Thanks, everyone for working to improve these documentation pages. 🎉

deivid-rodriguez:
Thanks for the review @olleolleolle and for getting this started @vog!

PR #518, opened 2020-04-28
Freeze Ruby version on CI and use it in cache key.

simi:


olleolleolle:
@simi I have commit bit, should I merge this right away?

simi:
should be ok, it makes CI green again (not sure if this was really the problem, but at least it wipes old cache which is probably the reason)

olleolleolle:
The result was: the build is now green.

PR #526, opened 2020-05-31
Link to the correct bundler issue tracker

bronzdoc:
Point links to the correct bundler issue tracker
closes rubygems/rubygems#3642

### DIFF
--- a/source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
+++ b/source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
@@ -427,9 +427,9 @@ encountering an issue, go to the [creating an issue][creating-an-issue] section 
 
 If none of these instructions fixed the problem, the next step is to open an issue.
 
-(Create an issue in the [RubyGems issue tracker](https://github.com/rubygems/rubygems/issues)
+(Create an issue in the [RubyGems issue tracker](https://github.com/rubygems/rubygems/issues/new?labels=Rubygems&template=rubygems-related-issue.md)
 if your error came from `gem install`. If it came from `bundle install`, create an issue in
-the [Bundler issue tracker](https://github.com/rubygems/rubygems/issues).)
+the [Bundler issue tracker](https://github.com/rubygems/rubygems/issues/new?labels=Bundler&template=bundler-related-issue.md).)
 
 Please include:
 - The output from running `gem env`:
--- a/source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
+++ b/source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
@@ -213,7 +213,7 @@ You're done! Run `gem --version` to verify that you are using the latest version
 ### Updating System Clock
 [update-system-clock]: #updating-system-clock
 
-If your system clock is set to a time in the past or future, your machine will not be able to establish a secure connection to RubyGems.org. To resolve the issue, you will need to set your system clock to the current time. In Linux, you can update the system clock by running `sudo ntpdate ntp.ubuntu.com`. 
+If your system clock is set to a time in the past or future, your machine will not be able to establish a secure connection to RubyGems.org. To resolve the issue, you will need to set your system clock to the current time. In Linux, you can update the system clock by running `sudo ntpdate ntp.ubuntu.com`.
 
 Here are other possible solutions for updating a system clock:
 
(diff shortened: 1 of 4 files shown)


source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
```diff
@@ -213,7 +213,7 @@ You're done! Run `gem --version` to verify that you are using the latest version
 ### Updating System Clock
 [update-system-clock]: #updating-system-clock
 
-If your system clock is set to a time in the past or future, your machine will not be able to establish a secure connection to RubyGems.org. To resolve the issue, you will need to set your system clock to the current time. In Linux, you can update the system clock by running `sudo ntpdate ntp.ubuntu.com`. 
+If your system clock is set to a time in the past or future, your machine will not be able to establish a secure connection to RubyGems.org. To resolve the issue, you will need to set your system clock to the current time. In Linux, you can update the system clock by running `sudo ntpdate ntp.ubuntu.com`.
```
deivid-rodriguez, line 216: We could add some markdown linter and be done with these unrelated whitespace changes. Anyways, no big deal, and not something to be fixed here.

source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
```diff
@@ -429,7 +429,7 @@ If none of these instructions fixed the problem, the next step is to open an iss
 
 (Create an issue in the [RubyGems issue tracker](https://github.com/rubygems/rubygems/issues)
 if your error came from `gem install`. If it came from `bundle install`, create an issue in
-the [Bundler issue tracker](https://github.com/bundler/bundler).)
+the [Bundler issue tracker](https://github.com/rubygems/rubygems/issues).)
```
deivid-rodriguez, line 432: We're linking to the same place twice. Better to reword the whole sentence to just "Create an issue in the [RubyGems issue tracker](https://github.com/rubygems/rubygems/issues)".

source/v2.0/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
```diff
@@ -429,7 +429,7 @@ If none of these instructions fixed the problem, the next step is to open an iss
 
 (Create an issue in the [RubyGems issue tracker](https://github.com/rubygems/rubygems/issues)
 if your error came from `gem install`. If it came from `bundle install`, create an issue in
-the [Bundler issue tracker](https://github.com/bundler/bundler).)
+the [Bundler issue tracker](https://github.com/rubygems/rubygems/issues).)
```
deivid-rodriguez, line 432: Same here, since the repos are merged, we can link to the merged repository just once. Or alternatively, change both links to use the proper template like: https://github.com/rubygems/rubygems/issues/new?labels=Bundler&template=bundler-related-issue.md, or something with rubygems.

source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
bronzdoc: Ah sorry, this is something that my editor does every time I save a file

source/v1.16/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
bronzdoc: Sounds good.

My "concern" is that people might get confused about it? maybe we should do it this way so people get familiar with the bundler/rubygems merge and then update the documentation to just point to the rubygems issue tracker?

deivid-rodriguez: Makes sense but in any case we should use a properly parameterized URL so that they are taken to the proper issue template, right?

bronzdoc: 🤔  I'm not understanding, we are leading them to the general issue tracker if they open an issue they will choose what issue to create there rubygems or bundler

deivid-rodriguez: In my opinion, there's two options:

* "Merge" this sentence to include a single link to the main tracker, so that once there they can change the template for either rubygems or bundler there. This is my intial suggestion.

* Keep the sentence as it is, but in that case, the users will be already making their choice of project when they click one link or the other, so we can use links like https://github.com/rubygems/rubygems/issues/new?labels=Bundler&template=bundler-related-issue.md or https://github.com/rubygems/rubygems/issues/new?labels=Rubygems&template=rubygems-related-issue.md. This is another alternative that makes sense to me.

I believe both options make sense and get users familiar with the merge.

deivid-rodriguez: Revisiting this after a while. What do you think about my last suggestion @bronzdoc?

bronzdoc: I went for option number two. Let me know if that is ok @deivid-rodriguez :)

PR #528, opened 2020-06-04
Highlighted a key precondition to use the Local Git Repository feature.

jbrains:
The problem was that I couldn't easily glean a key precondition to using the Local Git Repository feature, which resulted in spending several minutes trying to use the feature before realizing that I was doing it incorrectly.

My diagnosis was that a key piece of information was buried a few paragraphs down and difficult to spot until I read the documentation a few times.

My fix draws the reader's attention to this key piece of information.

I chose this fix because it seems like the smallest change to the existing documentation that could possibly provide value to the reader, helping them avoid my mistake.


jbrains:
Tack själv!

PR #530, opened 2020-07-20
Add documentation and release notes for bundler 2.1

deivid-rodriguez:

The problem was we never updated the documentation site for the 2.1 release.

My diagnosis was that we should do it, because users keep understandably referring to outdated docs, but they are the latest we have available.

My fix is:

* Generate a new version of the documentation for bundler 2.1
* Add a "What's new" page which is basically a transcription of https://github.com/rubygems/rubygems/blob/a0a7b1c169f063abd7a24b85dde2c604cf864501/bundler/UPGRADING.md, but with the format adapted to haml.
* Mention that we didn't publish a blog post for this version.

I chose this fix because using the current working infraestructure for building docs was the quickest way to fix this. In the future we want to either:

* Archive old versions of the docs.
* Stop publishing a new whole version of the whole doc site for every minor version. 

Closes https://github.com/rubygems/rubygems/issues/3599.


deivid-rodriguez:
I checked in the test deployment that [`bundle config` documentation is now up to date](https://bundler-site-build-new--ocyxm6.herokuapp.com/v2.1/man/bundle-config.1.html), including the changes made in https://github.com/rubygems/bundler/pull/7242, which should fix the problem raised in https://github.com/rubygems/rubygems/issues/3842.

deivid-rodriguez:
I could also write a small blog post, but I'm not sure it makes much sense at this point :sweat_smile:.

deivid-rodriguez:
If I recall correctly, there's no manual step here for deplying this, right? Merging this PR will do the trick?

deivid-rodriguez:
Thanks for your reviews, by the way!

deivid-rodriguez:
I'll merge this now and see how it goes!

PR #533, opened 2020-09-11
therbendo

therbendo:
Fixes #Thanks so much for the contribution!

The problem was...

My diagnosis was...

My fix...

I chose this fix because...


duckinator:
The `staging` branch has only ever been used by bundlerbot, and is not supposed to be updated manually. Thus, this PR serves no purpose. Closing.

PR #534, opened 2020-10-28
Plugin docs: mention uninstall & introduce BundlerError and explain how Bundler rescues those

jaredbeck:

The plugin guide did not cover these two topics.

See the docs I've written.

Wrote docs.

I'm not aware of other options.

### DIFF
--- a/source/v2.1/guides/bundler_plugins.html.haml
+++ b/source/v2.1/guides/bundler_plugins.html.haml
@@ -158,7 +158,7 @@ title: How to write a Bundler plugin
       %h4#raising_errors
         Raising Errors
       %p
-        If something goes wrong, your plugins should raise a `BundlerError`. It's not recommended to raise e.g. `Exception` in a plugin, because that will cause bundler to print its own bug report template, asking users to report the bug to bundler itself.
+        If something goes wrong, your plugins should raise a `BundlerError`. It's not recommended to raise e.g. `Exception` in a plugin, because that will cause Bundler to print its own bug report template, asking users to report the bug to Bundler itself.
       %p
         To see in detail how bundler rescues errors, check out `bundler/friendly_errors.rb`.
 
--- a/source/v2.1/guides/bundler_plugins.html.haml
+++ b/source/v2.1/guides/bundler_plugins.html.haml
@@ -122,6 +122,10 @@ title: How to write a Bundler plugin
           # The exec method will be called with the `command` and the `args`.
           # This is where you should handle all logic and functionality
           def exec(command, args)
+            if args.empty?
+              # Using BundlerError in plugins is recommended. See below.
+              raise BundlerError, 'My plugin requires arguments'
+            end
             puts "You called " + command + " with args: " + args.inspect
           end
         end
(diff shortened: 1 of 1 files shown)


source/v2.1/guides/bundler_plugins.html.haml
```diff
@@ -62,6 +62,9 @@ title: How to write a Bundler plugin
       %p
         Run <code>bundler plugin help install</code> for more details help and instructions on installing from Git.
 
+      %p
+        In Bundler 2.2.0, you can uninstall with <code>bundler plugin uninstall <gem_name></code>.
```
olleolleolle, line 66: Question: Does the word between `<` and `>` turn into visible text? I am looking at the web page: note that the `<gem_name>` is invisible.

bild


Workaround/question: Would it be easier to name an existing plugin as an example? `bundler plugin uninstall florp`

Another piece of the documentation of plugins mentions the gem name `copycat`. That sidesteps the `<` and `>` issue.

olleolleolle (commented):
Thanks for adding documentation text - that's super.

I found an area we can improve before merging. WDYT?

source/v2.1/guides/bundler_plugins.html.haml
olleolleolle: Oh, perhaps `bundler plugin uninstall bundler-florp`

jaredbeck: Hi Olle. Good catch. Any of these are fine with me.

olleolleolle: Suggetion: You make a pick, and change to what seems best – you're a plugin author, after all.

jaredbeck: > You make a pick, and change to what seems best

I decided to go with just `gem_name` (no brackets), partly because it was already used above.

jaredbeck:
Would you be interested in a separate PR adding an "API reference page", like `v2.1/bundle_plugin.html.haml`, for the `plugin` commands?

olleolleolle:
@jaredbeck It could definitely help!

source/v2.1/guides/bundler_plugins.html.haml
```diff
@@ -148,6 +155,13 @@ title: How to write a Bundler plugin
         %li
           The class defines the instance method <code>exec(command_name, args)</code>
 
+      %h4#raising_errors
+        Raising Errors
+      %p
+        If something goes wrong, your plugins should raise a `BundlerError`. It's not recommended to raise e.g. `Exception` in a plugin, because that will cause bundler to print its own bug report template, asking users to report the bug to bundler itself.
```
olleolleolle, line 161:
```suggestion
        If something goes wrong, your plugins should raise a `BundlerError`. It's not recommended to raise e.g. `Exception` in a plugin, because that will cause Bundler to print its own bug report template, asking users to report the bug to Bundler itself.
```

How do you feel about this styling of the Bundler name, here?

olleolleolle (approved):
This fixes a visual bug rendering the install instructions in Markdown.

It also explains more about exceptions, and points out which Bundler versions support `uninstall`.

🎉 Thanks, @jaredbeck!

jaredbeck:
>> Would you be interested in a separate PR adding an "API reference page", like v2.1/bundle_plugin.html.haml, for the plugin commands?
> It could definitely help!

https://github.com/rubygems/bundler-site/pull/535

PR #535, opened 2020-10-30
Reference page for plugin commands

jaredbeck:

Page missing

Added page

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -164,7 +164,7 @@ GEM
     rb-fsevent (0.10.3)
     rb-inotify (0.10.0)
       ffi (~> 1.0)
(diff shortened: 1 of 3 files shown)


olleolleolle (approved):
minor detail: we refer to "checkout" (a noun, but also the name of a Git command) and "check out" (a verb). We should choose only one way to refer to that. I suggest we pick the verb form. 

Adding this page raises the status of the plugin system, and increases plugin developers' chances.

jaredbeck:
> minor detail: we refer to "checkout" (a noun, but also the name of a Git command) and "check out" (a verb). We should choose only one way to refer to that. I suggest we pick the verb form.

I don't have a preference on this topic. That text is copied from `bundle plugin help install`.

Issue #536, opened 2020-11-05
/commands.html and dead links from it


tnir:
There are many dead links from https://bundler.io/commands.html:

- https://bundler.io/bundle_config.html
- https://bundler.io/bundle_exec.html
- https://bundler.io/bundle_gem.html
- https://bundler.io/bundle_package.html
- https://bundler.io/bundle_platform.html
- https://bundler.io/bundle_update.html

`https://bundler.io/commands.html` is at the first in the search results from `https://www.google.com/search?q=bundler+cli` (as of 2020-11-05).

deivid-rodriguez:
Hei! Thanks for reporting this. I have some hesitations about what to do here, because `commands.html` seems like a legacy page no longer linked from anywhere inside the site, and it's also quite confusing because it's styling makes it very hard to match command and applicable options. So, maybe we should get rid of it and redirect to a better page?

tnir:
Until I have to access to https://www.google.com/search?q=bundler+cli, I had the same opinion like you. If permits, do you think I can use redirections all from top-level to `<current_version>/page.html`?

deivid-rodriguez:
What we've been doing lately to play nicer with google is to use `<link href='/man/bundle-config.1.html' rel='canonical'>` from `https://bundler.io/v2.1/man/bundle-config.1.html` an so on, so that google considers them as the same page. However, I'm now observing that we also add canonical refs from `https://bundler.io/v2.0/man/bundle-config.1.html`, which is incorrect since they have different content.

I'm not sure how this plays with https://github.com/rubygems/bundler-site/pull/537.

tnir:
@deivid-rodriguez Thanks for your reply. So do you prefer `/man/bundle-install.1.html` than `https://bundler.io/bundle_config.html`? This issue is different from any pages under `/v2.1/`, `/v2.0` as well as `/man`...

deivid-rodriguez:
I was thinking about redirecting `/commands.html` to some better page (https://bundler.io/docs.html I think would be nice), and I don't think we need to do anything else? Pages like `https://bundler.io/bundle_config.html` don't exist and that seems fine?

tnir:
https://bundler.io/commands.html is still at the second place in the search results from https://www.google.com/search?q=bundler+cli (as of 2022-03-14).

deivid-rodriguez:
Thanks for the ping, I'll try fix this as I suggested previously.

deivid-rodriguez:
[...]

PR #537, opened 2020-11-05
Copy the latest command references to top-level

tnir:

`https://bundler.io/bundle_config.html` would be available instead of `Page not found` (see #536).

My diagnosis is to face `404 page not found` for `https://bundler.io/bundle_config.html`.

This PR fixes to add proxies from `/v2.1/<man-based-CLI-reference>.html` to `/<man-based-CLI-reference>.html`.

When v2.1 is the latest version of bundler, copy man-based docs to the top-level directory to avoid dead links as well as backward compatibility.

`/v2.1/bundle_config.html` (adapted from `bundler/bundler`) would be available at `/bundle_config.html` as well.

Keeping to use `bundler/bundler` instead of `bundler` dir of `rubygems/rubygems` is **NOT** a scope of this PR.

I chose this fix because the existing implementation uses `proxy` instead of  while I prefer redirections to avoid duplication.

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -63,21 +63,21 @@ Dir.glob("./source/localizable/#{config[:current_version]}/**/*").select{ |f| Fi
   proxy "#{country}/#{proxy_path}", page_path, locale: country.to_sym
   proxy proxy_path, page_path, locale: :en if country == 'en'
 end
+# Same for man-generated documentation
+# Ex: /v2.1/man/bundle-config.1.html.erb (adapted from bundler/bundler) available at /bundle_config.html
+Dir.glob("./source/#{config[:current_version]}/man/**/*").select{ |f| File.file?(f) }.each do |file_path|
+  handle_man_page(file_path) do |man_page_name, page_path|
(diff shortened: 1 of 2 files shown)


tnir:
cc @sailorhg @kruczjak (according to https://bundler.io/contributors.html ?)

tnir:
@sailorhg @kruczjak look not in charge of the site `bundler.io`, so @deivid-rodriguez is the corresponding person to this issue?

deivid-rodriguez:
Already looked into this and I'm not sure about it. Could you reply to https://github.com/rubygems/bundler-site/issues/536#issuecomment-722371146 and research about it?

tnir:
@deivid-rodriguez 's comments on #536 did not look relevant to this PR to me... Could you elaborate it???

simi:
Is this still needed?

tnir:
@simi https://bundler.io/bundle_config.html is still 404, but as these pages has less links than in 2020, let us close this issue 👍

PR #538, opened 2020-11-05
Remove an extra empty line from bundler console

tnir:

https://bundler.io/bundle_console.html

bundler io_bundle_console html(iPad Pro)

https://bundler.io/v2.1/bundle_console.html

bundler io_v2 1_bundle_console html(iPad Pro)

Removes an extra empty line from bundler console.
after

I chose this fix because formatting should be consistent with other pages.


deivid-rodriguez (approved):
Thanks!

PR #539, opened 2020-11-11
Fix typo

artur-beljajev:


olleolleolle (approved):
Thanks, @artur-beljajev! 

https://en.wiktionary.org/wiki/parallelly

(It's possible to use `in parallel` and side-step this, but today I learned, and I'm grateful for that.)

artur-beljajev:
Same here, @olleolleolle! We need to say thanks to Chrome, since it I just copy-pasted it to one of my commits and it highlighted it as being incorrect.

Issue #540, opened 2020-12-07
Fix SSL-related documentation on bundler.io


MartinKoerner:
This is an issue about the bundler.io documentation.
I can also prepare a pull request, if you hint the location of the source code of the website.

### Describe the problem as clearly as you can

The documentation of the recently changed Certificate is outdated:
https://bundler.io/v2.0/guides/rubygems_tls_ssl_troubleshooting_guide.html#automated-ssl-check (Chapter "Updating CA certificates")

1. The link "GlobalSignRootCA.pem" should refer to "https://github.com/rubygems/rubygems/blob/master/lib/rubygems/ssl_certs/rubygems.org/GlobalSignRootCA_R3.pem"
2. In "Step 3: Copy new trust certificate" it should be made clear to put the certificate in directory "...rubygems/ssl_certs/rubygems.org"

olleolleolle:
The new URL with the same properties as the existing one would be: 

`https://raw.githubusercontent.com/rubygems/rubygems/master/lib/rubygems/ssl_certs/rubygems.org/GlobalSignRootCA_R3.pem`

olleolleolle:
@MartinKoerner I made the URL change, and it's merged. If you want to clarify docs, go ahead!

olleolleolle:
@MartinKoerner If you want to offer a Pull Request about clarifying, go ahead!

MartinKoerner:
Sorry for the delay, I added #574

olleolleolle:
@MartinKoerner Thank you! :hearts: :yellow_heart: :green_heart: :blue_heart: :purple_heart:

PR #541, opened 2020-12-07
SSL troubleshooting: Update .pem link

olleolleolle:

The problem was that the linked .pem file 404d.

My diagnosis was taken from #540, and impl. here as a tiny change to the 2.1 document.

See #540

My fix is a wording change.

I chose this fix because: the other documents are of... historical value? Perhaps? Perhaps they should also change, now? What do _you_ think?


deivid-rodriguez (approved):
Thanks @olleolleolle, this looks good to me!

Not sure if we should change other versions, I guess it wouldn't hurt? Anyways, up to you!

olleolleolle:
Alright, I chose to go back and fix each reference to that link in the articles with the same name.

If green, I'll merge.

PR #557, opened 2021-02-17
Add a note about the improvement no longer being there

deivid-rodriguez:

We had to revert the change we bragged about :grimacing: 

Not really sure what the best way to handle this kind of situation is. I guess a note at the top of the blog post?

### DIFF
--- a/source/blog/2021-02-15-a-more-secure-bundler-we-fixed-our-source-priorities.html.markdown
+++ b/source/blog/2021-02-15-a-more-secure-bundler-we-fixed-our-source-priorities.html.markdown
@@ -9,6 +9,11 @@ category: security
 
 ---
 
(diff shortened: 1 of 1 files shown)


source/blog/2021-02-15-a-more-secure-bundler-we-fixed-our-source-priorities.html.markdown
```diff
@@ -9,6 +9,11 @@ category: security
 
 ---
 
+> __NOTE__: The part about the bundler release in this blog post is not currently
```
olleolleolle, line 12:
```suggestion
> __NOTE__: The part about the Bundler release in this blog post is not currently
```

deivid-rodriguez:
Thanks as always @olleolleolle!

PR #558, opened 2021-02-22
Deploy using GitHub Actions

deivid-rodriguez:

The problem was TravisCI is shutting down.

My diagnosis was we should move on.

My fix is to try to migrate to Github Actions.

I chose this fix because we are already using it in other places and works just fine.


indirect (approved):
looks good to me!

simi (approved):

olleolleolle:
@deivid-rodriguez There is an extra space just after `-iv`, perhaps something variable-like got "eaten by a grue"?

```
openssl aes-256-cbc -K  -iv  -in deploy_key.enc -out deploy_key -d
iv undefined
rake aborted!
Command failed with status (1): [openssl aes-256-cbc -K  -iv  -in deploy_ke...]
lib/tasks/ci/deploy.rake:15:in `configure_ssh_deploy_key'
lib/tasks/ci/deploy.rake:22:in `block (2 levels) in <top (required)>'
```

deivid-rodriguez:
The problem is that the decryption of the deployment key depends on some secret environment variables which are only set on TravisCI configuration. Since they are empty on Github Actions, we get this error.

We need to either find the original values of these secrets and migrate them to Github Actions secrets, or to create a fresh deployment key in the https://github.com/rubygems/bundler.github.io repository, encrypt it in the same way using Github Actions secrets, and overwrite the generated `deploy_key.enc` file.

deivid-rodriguez:
I tried this on this branch (removing the actual pushing to https://github.com/rubygems/bundler.github.io, and it seems to be working fine, so I guess we can merge and see how it goes.

deivid-rodriguez:
Just to further detail the steps I followed here:

* I created a new ssh-key and gitignored the private part.
* I encoded it using https://github.com/travis-ci/travis.rb, and noted the KEY and IV values used to encrypt it (I edited the gem sources directly to do this because by default the client uploads their values to travis-ci directly and can't be ever seen).
* I created the proper secrets inside this repository, and overwrote the generated `deploy_key.enc` in source control.
* I made sure the generated secrets are passed as environment variables to the deploy job. To properly implement this step, I removed the `ENCRYPTION_LABEL` environment variable previously used in travis-ci because it complicated passing secrets to the job (since env variable names were dynamic based on its value). I don't think it was providing any extra security.

deivid-rodriguez:
Ok let's try this.

PR #560, opened 2021-03-27
Update blog post note

deivid-rodriguez:

The problem was we made some improvements regarding dependency confusion, but didn't update this note.

My fix is to update the note to explain the current situation.

### DIFF
--- a/source/blog/2021-02-15-a-more-secure-bundler-we-fixed-our-source-priorities.html.markdown
+++ b/source/blog/2021-02-15-a-more-secure-bundler-we-fixed-our-source-priorities.html.markdown
@@ -9,10 +9,10 @@ category: security
 
 ---
 
-> __NOTE__: The part about the Bundler release in this blog post is not currently
-> relevant. We had to revert the change since it created breaking behavior for
-> our users. We'll remove this note once we release a properly working version
-> that implements this improvement.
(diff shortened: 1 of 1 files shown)


source/blog/2021-02-15-a-more-secure-bundler-we-fixed-our-source-priorities.html.markdown
```diff
@@ -9,10 +9,10 @@ category: security
 
 ---
 
-> __NOTE__: The part about the Bundler release in this blog post is not currently
-> relevant. We had to revert the change since it created breaking behavior for
-> our users. We'll remove this note once we release a properly working version
-> that implements this improvement.
+> __NOTE__: As of bundler 2.2.15, the part about the Bundler release in this
+> blog post is not fully relevant. Issue has been partially address for
```
olleolleolle, line 13: Adressed

olleolleolle (approved):

source/blog/2021-02-15-a-more-secure-bundler-we-fixed-our-source-priorities.html.markdown
deivid-rodriguez: Fixed, thanks @olleolleolle!

PR #561, opened 2021-03-28
Migrate from Sprockets to Webpack

deivid-rodriguez:

The problem was the the `middleman-sprockets` library is no longer maintained and it doesn't sound like it will be upgraded to support sprockets 4 in the near future. That holds other upgrades like rack, which currently shows a dependabot security alert.

My diagnosis was that the best move forward would be to migrate to webpack.

My fix is to migrate the app to use webpack. It seems to work fine in development but might need some more tweaks for production.


deivid-rodriguez:
As a cleanup prior to the migration I removed the JS for:

* https://www.seedanddew.com/. Because this service seems broken now, so we're probably not using it.
* Analytics file, because it seems to be failing to load in production so I don't think we're using it either.

I'd like confirmation that we're not really using these.

deivid-rodriguez:
Closing to force the review app to be recreated, I think it's needed to pick up my ENV changes.

PR #562, opened 2021-03-28
Migrate from Sprockets to Webpack (v2)

deivid-rodriguez:
This is a dup of #561.

The problem was the the `middleman-sprockets` library is no longer maintained and it doesn't sound like it will be upgraded to support sprockets 4 in the near future. That holds other upgrades like rack, which currently shows a dependabot security alert.

My diagnosis was that the best move forward would be to migrate to webpack.

My fix is to migrate the app to use webpack. It seems to work fine in development but might need some more tweaks for production.


deivid-rodriguez:
This PR still doesn't allow us to move on to a non vulnerable `rack` version (we should upgrade to `middleman` master for that), but unpins several other dependencies, so it moves in that direction.

In addition to that, the `middleman` project seems semi-abandoned, so we might want to consider alternatives. In that case, this PR also moves things forward in that direction by removing several middleman plugins.

So, since it has no user visible changes and moves us forward towards a better maintained stack, I'll be merging it shortly.

deivid-rodriguez:
Actually, a couple of days ago, the requirement on rack in middleman's 4.x stable branch was relaxed: https://github.com/middleman/middleman/pull/2460. So we can upgrade now :)

simi:
:partying_face:

PR #567, opened 2021-04-14
Fix a broken link on plugin page

okuramasafumi:
The link should be `.html`, not `.html.haml`

The problem was the link in https://bundler.io/v2.2/bundle_plugin.html was broken.

My diagnosis was the link is `.html.haml`.

My fix is to remove `.haml` part from the link.

I chose this fix because other working links end with `.html`, not `.html.haml`.

### DIFF
--- a/source/v2.1/bundle_plugin.html.haml
+++ b/source/v2.1/bundle_plugin.html.haml
@@ -8,7 +8,7 @@
       %ul
         %li
           See also
-          = link_to 'How to write a Bundler plugin', './guides/bundler_plugins.html.haml'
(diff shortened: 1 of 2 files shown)


olleolleolle (approved):
💯

olleolleolle:
https://github.com/rubygems/bundler-site/blob/master/source/v2.1/bundle_plugin.html.haml Can you fix this one, as well?

okuramasafumi:
@olleolleolle Fixed that too!

olleolleolle (approved):
@okuramasafumi Ah, thanks! Merging!

PR #573, opened 2021-05-18
Update the Contributing link to the new URI

olleolleolle:
I made the change in the index file and in the v2.2

The problem was that the links were pointing to the old repo.

My diagnosis was we needed to update them to current ones.

My fix is to point to the README in the rubygems/rubygems repo.

I chose this fix – fixing the source file and the latest generated version's HTML, since it was quick to do, and not too onerous to review.

Also: this is a focused change, which does not do All The Things.

### DIFF
--- a/source/localizable/index.html.haml
+++ b/source/localizable/index.html.haml
@@ -68,13 +68,13 @@
     .col-md-12
       %h2#get-involved Get involved
       .large-font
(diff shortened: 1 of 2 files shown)


source/localizable/index.html.haml
```diff
@@ -68,13 +68,13 @@
...
 
       .contents
         .buttons
           = link_to 'Code of Conduct', '/conduct.html', class: 'btn btn-primary'
           = link_to 'Join the Bundler Slack', 'http://slack.bundler.io', class: 'btn btn-primary'
-          = link_to 'Contributing guide', 'https://github.com/bundler/bundler/blob/master/doc/contributing/README.md', class: 'btn btn-primary'
+          = link_to 'Contributing guide', 'https://github.com/rubygems/rubygems/tree/master/bundler/doc/contributing/README.md', class: 'btn btn-primary'
```
deivid-rodriguez, line 77: `tree` redirects to `blob`, so I guess better keeping "blob"?

olleolleolle:
@deivid-rodriguez I got antsy, sorry, and I hit the merge button, I didn't mean to, I was investigating why the GitHub Actions were showing "Skipped" and a red marker.

deivid-rodriguez:
No worrries, and thanks for your work! :heart:

PR #581, opened 2021-07-11
Rename git.html.haml to git.html.haml.css

Iceymann18777:

The problem was...

My diagnosis was...

My fix...

I chose this fix because...


simi:
:information_source: user reported

PR #584, opened 2021-08-05
`npm update tar`

deivid-rodriguez:

A github security alert.

Bump tar.


olleolleolle (approved):
Looks 👍 !

PR #588, opened 2021-12-08
Try to fix some security alerts

deivid-rodriguez:

The problem was GitHub security alerts.

Try bump the proper dependencies.


indirect (approved):
The good news is that the Bundler site is 100% static and served by Fastly / GitHub Pages, so it should be impossible for any of these security alerts to impact the live site. Thanks for updating things though!

deivid-rodriguez:
Thank you! To be honest, it was just an excuse to play around a bit with `npm`. `npm audit fix` is pretty cool!

I created #589 with something I forgot here, by the way.

PR #589, opened 2021-12-09
Bump ruby in deploy job too

deivid-rodriguez:

I forgot to bump ruby in the deploy job at #588.

My fix is to bump it.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


olleolleolle:
If we use "3.0" as the description of the Ruby version we want, the ruby/setup-ruby Action will pick its newest patch release and use that. Perhaps this is a change that we can use in scripts like these?

deivid-rodriguez:
I think heroku enforces an exact ruby version via Gemfile and lockfile, so I think this needs to match the Gemile and lockfile. I also like the determinism of exact versions. But if your way works, I don't mind trying.

PR #590, opened 2021-12-09
build: Document bundler-cache, and rely on it

olleolleolle:

The problem was very minor: CI looked like it did 2 `bundle install` actions.

My diagnosis was that we could use ruby/setup-ruby's `bundler-cache: true` and only that.

My fix was to remove the extra bundle install step. And then to document the `bundler-cache: true` setting in all places used.

I chose this fix because I wanted speed and clarity configuration.


deivid-rodriguez (approved):
:muscle:

olleolleolle:
I hit merge, I hope that was the way to go.

PR #592, opened 2021-12-19
Remove outdated link for bundle command

sonalkr132:
The original owner most likely deleted their profile on github


indirect (approved):
thanks!

deivid-rodriguez:
I think the user actually "renamed" their profile (delete + create a new one, I guess), and the new home for the plugin is https://github.com/kddnewton/bundler-console?

indirect:
oh, good catch! we can always add that to the list 👍🏻

Boobie84:
> The original owner most likely deleted their profile on github

PR #595, opened 2022-01-23
Update some dependencies

deivid-rodriguez:

The problem was none. Just keeping things up to date.

`npm update` and others.


olleolleolle:
Good stuff removing that README section, and TIL about v2 of setup-node. Thanks!

simi (approved):

deivid-rodriguez:
Thanks! I will follow up with a PR to enable bundler 2.3 documentation.

PR #603, opened 2022-03-14
Remove versions prior to 1.12 from documentation

deivid-rodriguez:

The problem was that we document too many old versions nobody cares about.

A few problems derivated from this are:

* They all have broken CSS so they are not very easy to read. v1.12 is the first version with correct CSS.
* 1.12 still has some considerable usage, according to rubytogether.ecosystem.org, but usage of older versions seems negligible, and as a consequence, I don't think docs for those are useful.
* We get Heroku warnigns about slug size with every deploy, I think this might fix them.
* I want to get rid of a very old commands.html page, ranked very high on Google. The only internal links to it are from these versions.

My diagnosis was that we should get rid of these versions.

My fix is to remove all the things.


olleolleolle (approved):
For the very interested reader, perhaps archive.org helps with this "find an old page". 

Good fix!

deivid-rodriguez:
According to https://ecosystem.rubytogether.org/?unique=true, while 1.12 is still the most used version of Bundler 😞, no versions below 1.12 show up in top 7, so I believe we can safely assume nobody is that far behind.

simi (approved):

hsbt (approved):

PR #604, opened 2022-03-15
Remove old `commands.html` page

### LINKED ISSUE #536: /commands.html and dead links from it

deivid-rodriguez:

The problem was that a highly ranked `commands.html` page includes partially broken styles, dead links, and it's not linked from anywhere else on the site.

My diagnosis was that we should remove this legacy page.

My fix is to remove the legacy page and the legacy sidebar that linked to it, and redirect the url to more up to date yet similar content.

I chose this fix because this page is no really maintained any more. I could've fixed the broken links and worked on appearance, but this is just an old page before the 1.12 redesign, it should be removed.

Fixes #536.


olleolleolle (approved):
Perfect.

With a redirect, we keep all the Google juice.

tnir:
> With a redirect, we keep all the Google juice.

Maybe this was not perfect.. Follows up this in #741, but I think it is now too late to add additional redirection rules, so I just remove only the rule of Middleman there.

PR #615, opened 2022-06-22
Fix dead links on GitHub from man pages

### LINKED ISSUE #620: Many editable pages are dead
Out of 946 pages [1], there are 807 links to blob page on GitHub, but many of them (~46%) are dead.

```
{200=>430, 404=>377}
```

- [1] Found URLs from https://bundler.io/sitemap.xml

tnir:

Dead links to source Git happen on both for v2.3 and earlier man pages.

Example:

https://bundler.io/v2.3/man/bundle-install.1.html has a link entitled as "Edit this document on GitHub" to `https://github.com/rubygems/rubygems/blob/master/bundler/man/bundle-install.1.ronn`, which is dead.

This regression has happened since 2020-09-04 with https://github.com/rubygems/rubygems/pull/3921.
A solution for this problem was not included in #542.

Fixes the dead links on GitHub from man pages from the current version of Bundler (2.3 as of writing).

Removes the dead links on GitHub from man pages from the non-current version of Bundler (1.12 through 2.2 as of writing). Instead of that, the link to the latest version's man is added there.

~~An end-user (website reader) should be able to consider if they can improve older Bundler man pages like 1.12 etc.~~

As the legacy Bundler maintenance policy will be updated in https://github.com/rubygems/rubygems/issues/5647, we do not have to take care about something wrong in man pages in Bundler 2.2 or earlier.

Partially resolves #620

### DIFF
--- a/helpers/docs_helper.rb
+++ b/helpers/docs_helper.rb
@@ -11,23 +11,34 @@ module DocsHelper
   end
 
   def link_to_editable_version
+    editable = true
     path = current_page.file_descriptor.relative_path.to_s
     repo =
       if path.start_with?('doc/')
         path = "bundler/#{path}"
         'rubygems/rubygems'
-      elsif path =~ %r{\Av\d+\.\d+/man/(bundle[_-]|gemfile)}
-        path = "#{strip_version_from_url(path)}.ronn"
-        path = "bundler/#{path}"
+      elsif %r{\A(?<version>v\d+\.\d+)/man/(?<filename>(bundle[_-]|gemfile)[^/]*)\.html} =~ path
+        if version == current_version
+          path = "bundler/lib/bundler/man/#{filename}.ronn"
+        else
+          editable = false
+        end
         'rubygems/rubygems'
       else
         path = File.join 'source', path
         'rubygems/bundler-site'
       end
-    url = "https://github.com/#{repo}/blob/master/#{path}"
 
-    link_to('Edit this document on GitHub', url) +
-      ' if you caught an error or noticed something was missing.'
+    if editable
+      url = "https://github.com/#{repo}/blob/master/#{path}"
+      link_to('Edit this document on GitHub', url) +
+        ' if you caught an error or noticed something was missing.'
+    else
+      url = "/" + path.sub(version, current_version).sub(/\.html.*/, '.html')
+      'This document is obsolete. ' +
+        link_to('See the latest version of this document', url) +
+        ' if you caught an error or noticed something was missing, it may be fixed there.'
+    end
   end
 
   def path_exist?(page, version=nil)


helpers/docs_helper.rb
```diff
@@ -1,5 +1,18 @@
...
+
+  VERSION_TO_REF = {
+    "v1.12" => "bundler-v1.12.6",
+    "v1.13" => "bundler-v1.13.7",
+    "v1.14" => "bundler-v1.14.6",
+    "v1.15" => "bundler-v1.15.4",
+    "v1.16" => "bundler-v1.16.6",
+    "v1.17" => "bundler-v1.17.3",
+    "v2.0" => "bundler-v2.0.0.pre.3", # TODO "bundler-v2.0.2"
+    "v2.1" => "bundler-v2.1.0.pre.3", # TODO "bundler-v2.1.4"
```
tnir, line 12: A workaround is applied. https://github.com/rubygems/rubygems/issues/5646 will improve these lines.

deivid-rodriguez: These are from before unifying development at the RubyGems repo, I think it would be best to link to the tag in the old repo for old releases.

deivid-rodriguez: In any case, could we avoid showing the link at all on older versions, and just show it when browsing the latest one?

tnir: > I think it would be best to link to the tag in the old repo for old releases

The link this PR is fixing is a editable resource on GitHub, so rubygems/bundler is not helpful as it was already archived.

> In any case, could we avoid showing the link at all on older versions, and just show it when browsing the latest one?

Makes sense to some degree and I considered it on inception. But what are older versions of Bundler? No EOL/deprecation policy in Bundler as fa as I know...

tnir: > [...] what are older versions of Bundler?

Asked in https://github.com/rubygems/rubygems/issues/5647. I doubt only @deivid-rodriguez answer here or there 😀

deivid-rodriguez: Hi @tnir.

You're right that linking to an archived repository makes no sense. However, linking to a tag makes no sense either, because tags are not editable either. Note that while https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/man/bundle-install.1.ronn has the "edit button" enabled, https://github.com/rubygems/rubygems/blob/bundler-v2.2.33/bundler/lib/bundler/man/bundle-install.1.ronn has it disabled.

So in my opinion the best solution here is to:

* If not in the latest version docs, display something like 

> [See the latest version of this document](https://bundler.io/v2.3/man/bundle-install.1.html) if you caught an error or noticed something was missing, it may be fixed there.

* If on the latest version of docs, link to the "Edit page" in the main branch:

> [Edit this document on GitHub](https://github.com/rubygems/rubygems/edit/master/bundler/lib/bundler/man/bundle-install.1.ronn) if you caught an error or noticed something was missing.

How about that?

deivid-rodriguez: Great! I noticed that there's still one broken link here: https://bundler-site-tnir-fix-l-kiodqk.herokuapp.com/man/bundle-install.1.html. That page is linked from the "CLI docs" link in homepage and should have the same content as the v2.3 one, except for this link for some reason. Can you have a look?

tnir: I have noticed the problem you noticed before creating this PR, but this PR does not change the behavior for it intentionally. (see https://bundler.io/man/bundle-install.1.html)

I think `/man/bundle-install.1.html` should be redirect to `/v2.3/man/bundle-install.1.html` (if the current version is v2.3) Thoughts?

tnir: ^ Just filed as #618.

deivid-rodriguez: Oh yes, I also noticed that the problem exists before this PR, I just thought I would mention it in case you wanted to fix it here too. It's up to you, if you want to fix that too, you can do it here or in a separate PR.

A redirect looks good to me, but I would also update the CLI docs link to link to `/v2.3/man/bundle-install.1.html` (or whatever the latest version is) directly (without redirect).

deivid-rodriguez (commented):
Looks good, just a small comment on the name of a variable.

helpers/docs_helper.rb
```diff
@@ -11,23 +11,34 @@ def link_to_documentation(page, version=nil)
   end
 
   def link_to_editable_version
+    linkable = true
```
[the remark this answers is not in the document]
tnir: Good catch! At that time I considered the message, I planed not to link from older version... 😁  Fixed in the latest commit.

deivid-rodriguez (approved):
Thank you, looks great!

PR #617, opened 2022-06-23
Enforce no trailing whitespace

deivid-rodriguez:

Just some annoying trailing whitespace in Ruby code.

Disallow it through RuboCop.

It makes sure we don't reintroduce the distracting inconsistency.


tnir:
❤️

PR #624, opened 2022-07-04
Aggregate guides for 1.12-1.14 to 1.15

tnir:
### What was the end-user problem that led to this PR?

Guides for older Bundler versions (v1.12-v1.14) are still available via side nav bar and via search engines.

bundler io_v1 12_whats_new html(iPad Air)

### What was your diagnosis of the problem?

There was no (significant) change in these guides until #404 (which was for guides for v1.16).

### What is your fix for the problem, implemented in this PR?

All guides (in en and pl) in v1.12 through v1.14 are now redirected to v1.15 guides in each language.

#### `$ diff -u source/localizable/v{1.12,1.15}/guides/creating_gem.en.html.md`
```diff
$ diff -u source/localizable/v{1.12,1.15}/guides/creating_gem.en.html.md
--- source/localizable/v1.12/guides/creating_gem.en.html.md     2022-07-04 20:15:03.832843559 +0000
+++ source/localizable/v1.15/guides/creating_gem.en.html.md     2022-06-23 14:18:04.000000000 +0000
@@ -13,6 +13,13 @@
 our own gems? It's really easy to do this and Bundler provides a couple of things to help you 
 along this path.
 
+Related reading:
+
+* [Creating a gem with a C
+  extension](http://guides.rubygems.org/gems-with-extensions/)
+* [Common practices for developing
+  gems](http://guides.rubygems.org/patterns/)
+
 ## But first, why?
 
 Why should we create a gem? Can't we just throw in some code into our *other* library and use 
@@ -29,18 +36,21 @@
 we might not get the exact same output.  To check which version of bundler we currently have, 
 lets run the following command:
 
-    bundle -v
+    $ bundle -v
 
 We should see something close to `Bundler version 1.9.0`.  If necessary, we can update to the 
 newest version of Bundler by running `gem update bundler`.
 
 To begin to create a gem using Bundler, use the `bundle gem` command like this:
 
-    bundle gem foodie
+    $ bundle gem foodie
 
 We call our gem `foodie` because this gem is going to do a couple of things around food, such as 
 portraying them as either "Delicious!" or "Gross!". Stay tuned.
 
+For information on gem naming conventions, you can read the &ldquo;[Name Your
+Gem](http://guides.rubygems.org/name-your-gem/)&rdquo; guide on the RubyGems website.
+
 This command creates a [scaffold directory](https://github.com/radar/guides/tree/master/gem-scaffold/foodie)
 for our new gem and, if we have
[...]

PR #625, opened 2022-07-04
Aggregate localizable guides for 1.16-2.2 to 2.3

tnir:
Blocked by #624

While guides are versioned with URLs, they have not been changed since Bundler v1.16.

Guides were duplicated via `rake versions:create`.

- Existing duplicated codes are removed.
- New duplication will stop.

`https://bundler.io/v2.3/guides/creating_gem.html` should be now `https://bundler.io/guides/creating_gem.html` as we do not have to version this guide in fact.


source/partials/_guides_sidebar.haml
```diff
@@ -1,6 +1,9 @@
+- @guides = guides
 %ul
-  - guides.each do |page|
+  - @guides.each do |page|
     %li{class: current_guide?(page[:filename])}
       = link_to_guide(page)
     %li.separator
       %hr
+  - if @guides.size == 0
+    %li ❗There is any guide for this Bundler version any more. Check out guides for the latest version.
```
deivid-rodriguez, line 9: So if I understand correctly, now the sidebar with the list of guides is only displayed for 1.15 and 2.3, and the rest show this alert?

I actually think we can take this a bit further. How about:

* Make guides completely version independent. If the only difference between them is usage of modern hashes, I think we can add a note at the beginning that the guide does not work on old Rubies due to this, and that users of these Rubies should upgrade. From now on, if we introduced something version dependent to any guide, we can add a note about the version where the feature described was first introduced.

* Replace the sidebar for each "what's new page" to instead show the documentation sidebar for each version (the one [here](https://bundler.io/v2.3/man/bundle-install.1.html)).

* Keep the guide list only at https://bundler.io/docs.html, which is version independent, and redirect all versioned urls to the non-versioned urls.

How about that?

tnir: Totally agree wit you, especially first and second ones.

> Keep the guide list only at https://bundler.io/docs.html, which is version independent, and redirect all versioned urls to the non-versioned urls.

Per last one, it depends on traffic (page views) of `/docs.html`. Weeks ago I found https://bundler.io/doc/policies.html, which has no link from. If `/docs.html` has less traffic like `/doc/policies.html`, that would be a living dead...

deivid-rodriguez: Can you answer

> So if I understand correctly, now the sidebar with the list of guides is only displayed for 1.15 and 2.3, and the rest show this alert?

If this is correct, I think for now, before we implement the other steps, I'd rather keep the list of links but change them to be redirects to either the 1.15 guide or the 2.3 guide accordingly.

tnir: > So if I understand correctly, now the sidebar with the list of guides is only displayed for 1.15 and 2.3, and the rest show this alert?

Yes. But,

> [...] I'd rather keep the list of links but change them to be redirects to either the 1.15 guide or the 2.3 guide accordingly.

I do not think it is easy to do without an additional workaround for version conversion from 1.12-1.14 and 1.16-2.2 to 1.15 and 2.3.

config.rb
```diff
@@ -93,6 +93,14 @@
   end
 end
 
+# Redirect old localizable guides (v1.16-v2.2) to the latest version (v2.3) of guide compatible
+["", "pl/"].each do |lang|
+  %w[1.16 1.17 2.0 2.1 2.2].each do |version|
+    redirect "#{lang}v#{version}/guides/creating_gem.html", to: "#{lang}v2.3/guides/creating_gem.html"
+    redirect "#{lang}v#{version}/guides/using_bundler_in_applications.html", to: "#{lang}v2.3/using_bundler_in_applications.html"
```
deivid-rodriguez, line 100: This is introducing the same bug fixed by #642.

helpers/guides_helper.rb
```diff
@@ -4,7 +4,7 @@ module GuidesHelper
 
   def guides
     guides = Dir.glob("./source/#{current_visible_version}/guides/*")
-    target_version = [current_visible_version, "v1.15"].max
+    target_version = current_visible_version > "v1.15" ? current_version : "v1.15"
```
deivid-rodriguez, line 7: This is unnecessary, I guess it comes from the rebase.

deivid-rodriguez (commented):
It seems another step in the good direction, just left two comments.

config.rb
tnir, line 100: Analogy to #642, this should be:
```suggestion
    redirect "#{lang}v#{version}/guides/using_bundler_in_applications.html", to: "#{lang}v2.3/guides/using_bundler_in_applications.html"
```

helpers/guides_helper.rb
tnir: Nope, `current_visible_version` is the version to render in Middleman and `current_version` is always `v2.3` (as of now).

deivid-rodriguez: Oh, sorry, my bad, I misread the code. Thanks!

deivid-rodriguez (approved):
Thanks!

PR #641, opened 2022-07-06
Update Spanish guides to catch up with latest changes

deivid-rodriguez:

The problem was these guides were not updated in recent versions.

My fix to adapt the changes applied to the English guides to these translated guides, so they are available to at https://bundler.io/es/v2.3/guides/bundler_setup.html and https://bundler.io/es/v2.3/guides/bundler_setup.html
 and https://bundler.io/es/v2.3/guides/bundler_sharing.html.

```diff
--- source/localizable/v1.15/guides/bundler_setup.es.html.haml	2022-07-06 11:16:59.000000000 +0200
+++ source/localizable/v2.3/guides/bundler_setup.es.html.haml	2022-07-06 11:17:30.000000000 +0200
@@ -3,7 +3,7 @@
 ---
 
 .container.guide
-  %h2 Configurando Bundler
+  %h2 Cómo usar Bundler con Ruby
 
   .contents
     .bullet
@@ -96,7 +96,7 @@
           source 'https://rubygems.org'
 
           gem 'rails', '5.0.0'
-          gem 'rack-cache', :require => 'rack/cache'
+          gem 'rack-cache', require: 'rack/cache'
           gem 'nokogiri', '~> 1.4.2'
       .bullet
         .description
--- source/localizable/v1.15/guides/bundler_sharing.es.html.haml	2022-07-06 11:16:59.000000000 +0200
+++ source/localizable/v2.3/guides/bundler_sharing.es.html.haml	2022-07-06 11:17:31.000000000 +0200
@@ -3,7 +3,7 @@
 ---
 .container.guide
   %h2#sharing
-    Compartiendo su código
+    Cómo paquetizar y compartir código usando un Gemfile
   .contents
     .bullet
       .description
@@ -55,7 +55,7 @@
         En otras palabras, no va a adivinar cuales versiones de las dependencias debería
         instalar. En los ejemplos que hemos usado, aunque <code>rack-cache</code> declara una
         dependencia en <code>rack >= 0.4</code>, sabemos que definitivamente trabaja con <code>rack
-        1.2.1</code>. Aunque Rack lanzaría <code>rack 1.2.2</code>, bundler va a siempre instalar
-        <code>1.2.1</code>, la version exácta de la gema que sabemos que funciona. Esto
+        1.5.2</code>. Aunque Rack lanzaría <code>rack 1.5.3</code>, bundler va a siempre instalar
+        <code>1.5.2</code>, la version exácta de la gema que sabemos que funciona. Esto
         alivia una gran parte de la carga de mantenimiento de desarrolladores de aplicación, porque
         todas las máquinas siempre ejecutarán el mismo código de terceros.
```

I chose this fix because regardless of whether these guides are actually useful for anyone or well translated, I think this unblocks further improvements to unify guides and make them not dependent on the bundler version.

### DIFF
--- a/source/localizable/v2.3/guides/bundler_setup.es.html.haml
+++ b/source/localizable/v2.3/guides/bundler_setup.es.html.haml
@@ -1,0 +1,108 @@
+---
+title: Usando Bundler con Ruby
+---
+
+.container.guide
+  %h2 Cómo usar Bundler con Ruby
+
+  .contents
+    .bullet
+      .description
+        Configure la ruta de carga para que todas las dependencias en
+        su Gemfile puedan ser requerídas
+    :code
+      # lang: ruby
+      require 'rubygems'
+      require 'bundler/setup'
+      require 'nokogiri'
+
+    .bullet
+      .description
+        Agregue solamente gemas de grupos específicos
+        a la ruta de carga. Si quiere usar las gemas en el grupo
+        por defecto, inclúyalo
+      :code
+        # lang: ruby
+        require 'rubygems'
+        require 'bundler'
+        Bundler.setup(:default, :ci)
+        require 'nokogiri'
+      = link_to 'Aprende más: Grupos', './groups.html', class: 'btn btn-primary'
+
+    %h2 Compatibilidad
+
+    .contents
+      .bullet
+        .description
+          Tanto Ruby 2.0 como Rubygems 2.0 requieren Bundler 1.13 o superior. Si tiene dudas sobre la compatibilidad entre Bundler y su sistema, por favor chequee la lista de compatibilidad.
+        = link_to 'Aprende más: Compatibilidad', '/compatibility.html', class: 'btn btn-primary'
+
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Ok, let me check we're on the same page:

#### Your proposal

* Make 1.12-15 redirect to 1.15 version of guides.
* Make 1.16-2.3 redirect to 2.3 version of guides.
* So, keep two copies of each guide.

#### My proposal

* Make all guides redirect to the latest version of each guide.
* So, keep one copy of each guide.

I thought you said that if we go with my idea, we "lose" the Spanish guides because the were no updated so the old versions would be lost. That's why I updated them, so that we can go with my solution next.

deivid-rodriguez:
I think we can move this forward now, and I think after this we're ready to also remove the 1.15 versions of guides, and move every guide to urls that are not dependent on versions.

tnir:
Yes, we can do it either way and I was fine either way. I will do the next step on top of this 🚀

PR #646, opened 2022-07-07
npm install is required to setup dev env

tnir:

To run `npm install` in the development setup instruction is not included and the server looks running when `bundle exec middleman server` is executed even without webpack installed.

A novice contributor can set up the environment for development. `npx webpack` dismisses the warning.

Just show it in the instruction.

Documentation is now but automation would be the next step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ Begin by cloning the repository onto your location machine:
 Once complete prepare the dependencies by running:
 
     bundle install
+    npm install
 
 ## Basic Middleman Commands
 


deivid-rodriguez (commented):
I think we can get `npx` to auto-install them if needed by passing `npx --yes` in `package.json` commands?

tnir:
> I think we can get npx to auto-install them if needed by passing npx --yes in package.json commands?

I do not think this way is recommended as it is considered as insecure to run `npx` without any confirmation... It looks like `npx webpack` was introduced in #562, but is it common? I have never seen it.

I think we can remove as follows:
```diff
diff --git a/package.json b/package.json
index af7cc1f..7b5e075 100644
--- a/package.json
+++ b/package.json
@@ -9,8 +9,8 @@
   "private": true,
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "start": "npx webpack -d source-map --progress --color",
-    "build": "npx webpack --bail --mode production"
+    "start": "webpack -d source-map --progress --color",
+    "build": "webpack --bail --mode production"
   },
   "repository": {
     "type": "git",
```
and then it would cause an error when `npm install` is missed:
```
$ bundle exec middleman server
== The Middleman is loading
== Executing: `npm run start`
sh: 1: webpack: not found
== External: > bundler-site@1.0.0 start
== External: > webpack -d source-map --progress --color
== External: Command failed with non-zero exit status
```
, which looks simple to me.

With the current settings, the error is dismissed due to `npx`:

```
$ bundle exec middleman server
== The Middleman is loading
== Executing: `npm run start`
== External: > bundler-site@1.0.0 start
== External: > npx webpack -d source-map --progress --color
== External: Need to install the following packages:
== External:   webpack
== Locales: en, es, pl (Default en)
== Blog Sources: /blog/{year}-{month}-{day}-{title}.html (:prefix + :sources)
== View your site at "http://localhost:4567", "http://127.0.0.1:4567"
== Inspect your site configuration at "http://localhost:4567/__middleman", "http://127.0.0.1:4567/__middleman"
```

deivid-rodriguez:
I added it because it was the closest equivalent to `bundle exec` that I could find. My intention was to make sure we run a version of `webpack` that we know works for us.

PR #647, opened 2022-07-08
Remove logos formerly used for sponsor showcase

tnir:

Intellectual properties (logos) from the (past) (direct) sponsors remain in the repo.

It would be a problem to have unused logos. The sponsor showcase itself was removed from the site in 2021 in 063784d2 (as part of #558).

Removes logos formerly used for sponsor showcase (at `/sponsors.html`).

We do not need them all. We can add one(s) when the project has (direct?) sponsor(s).


tnir:
@deivid-rodriguez Could you check if we can remove these image files previously used for sponsors? (I could not find anything else about sponsors in:)
- #558

deivid-rodriguez (approved):
I don't think we're using these anymore indeed, so seems fine!

PR #648, opened 2022-07-08
Add .gitpod.yml to make dev env easier

tnir:

The contribution guide has not been coded so hard for a new contributor to get started to contributions to the repository.

Gitpod (cloud edition) is helpful since everyone can use a workspace on it for 50 hours / months.

Added `.gitpod.yml` but there is no pre-built image including Ruby 3.1 and Node.js so we need to prepare it by ourselves.

#### How to test
1. Visit `https://github.com/tnir/bundler-site/tree/gitpod` and click a button to launch a Gitpod workspace for this.
2. First build will take some time (5 minutes and more).

Added a button to launch from README.md for novice contributors. I can do it for GitHub Codespaces as well in another PR.

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,4 @@
+FROM gitpod/workspace-ruby-3.1
+
+RUN curl -fsSL https://deb.nodesource.com/setup_16.x | sudo -E bash - && \
+  sudo apt-get install -y nodejs
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,10 @@
+image:
+  file: .gitpod.Dockerfile
+tasks:
+  - init: |
+      curl -fsSL https://deb.nodesource.com/setup_16.x | sudo -E bash -
+      sudo apt install -y nodejs
+      bundle config set path vendor/bundle
+      bundle install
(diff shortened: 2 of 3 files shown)


deivid-rodriguez:
I get an error page that says "no referrer found" when I click on the README button in your fork.

tnir:
😭 
Maybe you can fork to do so or I can be a member of https://github.com/rubygems/bundler-site. Which do you prefer?

deivid-rodriguez:
Oh, I see what happened here. I'm using Safari and the "from-referrer" trick does not seem to work on Safari. I understand that the URL using referrer would work on forks too, but maybe it's better to start with the "standard URL" that hardcodes our own path and see if something cares about clicking the button from a fork?

tnir:
Fixed it. You can test to open a workspace from `https://gitpod.io/#https://github.com/rubygems/bundler-site` while you cannot test how valid the `.gitpod.yml` is. You should test the validity of the `.gitpod.yml` **after you merged this to the default branch** if you want to check if this works. I have tested this `.gitpod.yml` several times and so I think it is tested enough.

deivid-rodriguez (approved):
Ok, if you have tested it it seems fine :)

PR #649, opened 2022-07-09
Replace rocket hashes with Ruby 1.9 ones

tnir:

Rocket hashes are used in Ruby code.

They can be replaced with new ones.

Replace rocket hashes in Ruby code with Ruby 1.9 ones.

Simplified code would give a better contributor experience.

No other cops are not enabled in this PR as per the title.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,3 +3,7 @@ AllCops:
 
 Layout/TrailingWhitespace:
   Enabled: true
+
+Style/HashSyntax:
+  EnforcedStyle: ruby19
+  EnforcedShorthandSyntax: never
--- a/Rakefile
+++ b/Rakefile
@@ -1,11 +1,11 @@
 Dir.glob("lib/tasks/**/*.rake").each { |r| load r }
 
 desc "Build the static site"
-task :build => ["contributors:update", :repo_pages, :man] do
+task build: ["contributors:update", :repo_pages, :man] do
   sh "middleman build --clean --verbose"
 end
 
 # Heroku runs assets:precompile during deploys
 namespace :assets do
-  task :precompile => :build
+  task precompile: :build
 end
--- a/config.ru
+++ b/config.ru
@@ -29,7 +29,7 @@ module Rack
 end
 
 use Rack::Deflater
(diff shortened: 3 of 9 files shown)


deivid-rodriguez (commented):
Can you enable the rule in `.rubocop.yml` and let RuboCop auto-correct everything?

deivid-rodriguez:
Why only rake tasks?

.rubocop.yml
```diff
@@ -3,3 +3,7 @@ AllCops:
 
 Layout/TrailingWhitespace:
   Enabled: true
+
+Style/HashSyntax:
+  EnforcedStyle: ruby19
+  EnforcedShorthandSyntax: never
```
deivid-rodriguez, line 9: I didn't know about this new shorthand syntax feature. It looks cool, why disable its use?

Rakefile
```diff
@@ -1,11 +1,11 @@
 Dir.glob("lib/tasks/**/*.rake").each { |r| load r }
 
 desc "Build the static site"
-task :build => ["contributors:update", :repo_pages, :man] do
+task build: %w[contributors:update repo_pages man] do
```
deivid-rodriguez, line 4: Can we also enable the cop to enforce `%w[]` for Arrays of Strings?

.rubocop.yml
tnir: Existing codes are all not in shorthand, so I just wanted to minimize the changes.

Rakefile
tnir: We can do it in another PR.

tnir: Or, I can revert this part to the following in this PR:

```suggestion
task build: ["contributors:update", :repo_pages, :man] do
```

tnir:
> Why only rake tasks?

I did not have a plan to change or consider other codes than Rake tasks.

Rakefile
tnir: Reverted.

tnir:
Back to you.

hsbt (approved):

hsbt:
Thanks!

deivid-rodriguez:
Thanks ❤️ !

Issue #653, opened 2022-07-11
Convert jQuery code with vanilla


tnir:
- [x] #654 (anchors.js, commands_layout.js)
- [x] #664 (search_arrows.js)
- [x] #665 (search.js)

PR #654, opened 2022-07-11
Remove jQuery dependency from commands layout

tnir:

- See also #653

Just follow the instruction from AnchorJS: https://github.com/bryanbraun/anchorjs/blob/master/README.md

HTMLs for command references were done from ones for search with Lunr.

### DIFF
--- a/assets/javascripts/commands_layout.js
+++ b/assets/javascripts/commands_layout.js
@@ -1,13 +1,11 @@
-import $ from 'jquery'
 import AnchorJS from 'anchor-js';
 
 const anchors = new AnchorJS();
 
-$('.version-selects').change(function(e) {
-  document.location.href = $(e.target).find(":selected").val();
+const el = document.querySelector('.version-selects');
+el.addEventListener('input', function (e) {
+  document.location.href = e.target.value;
 });
 
-$(document).ready(function() {
-  anchors.add('#page-content-wrapper h1, #page-content-wrapper h2, #page-content-wrapper h3, ' +
-    '#page-content-wrapper h4, #page-content-wrapper h5');
-});
+anchors.add('#page-content-wrapper h1, #page-content-wrapper h2, #page-content-wrapper h3, ' +
+  '#page-content-wrapper h4, #page-content-wrapper h5');
--- a/assets/javascripts/anchors.js
+++ b/assets/javascripts/anchors.js
@@ -1,8 +1,4 @@
-import $ from 'jquery'
 import AnchorJS from 'anchor-js';
 
 const anchors = new AnchorJS();
-
-$(document).ready(function() {
-  anchors.add();
-});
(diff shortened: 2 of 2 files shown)


olleolleolle:
Looks grand. Nice removal.

assets/javascripts/commands_layout.js
```diff
@@ -1,13 +1,11 @@
...
+el.addEventListener('input', function (e) {
+  document.location.href = e.target.value;
 });
 
-$(document).ready(function() {
-  anchors.add('#page-content-wrapper h1, #page-content-wrapper h2, #page-content-wrapper h3, ' +
-    '#page-content-wrapper h4, #page-content-wrapper h5');
-});
+anchors.add('#page-content-wrapper h1, #page-content-wrapper h2, #page-content-wrapper h3, ' +
+  '#page-content-wrapper h4, #page-content-wrapper h5');
```
olleolleolle, line 11: Minor: Perhaps we can use some formatting spaces to show that these strings are _very_ alike? E.g. one selector on each line, or something. Still as a selector, but formatted for easy scanning and understanding.

```suggestion
anchors.add(
  '#page-content-wrapper h1, ' +
  '#page-content-wrapper h2, ' + 
  '#page-content-wrapper h3, ' +
  '#page-content-wrapper h4, ' + 
  '#page-content-wrapper h5'
);
```

deivid-rodriguez (approved):
This looks great to me.

If @tnir wants to address @olleolleolle's comment seems fine. If postponing @olleolleolle's suggestion to a different PR to minimize changes is preferred, that's also fine with me!

assets/javascripts/commands_layout.js
olleolleolle: This is unrelated to the PR, but I just thought of this change while reading. Not important that it goes in with this PR.

tnir:
@deivid-rodriguez Thanks. I defer it to another PR.

assets/javascripts/commands_layout.js
tnir: Thanks again @olleolleolle 😍. Good point. I guess there is no `h5` (or `h1`) in `#page-content-wrapper`, so we should remove these.

```js
anchors.add('#page-content-wrapper h2, #page-content-wrapper h3, #page-content-wrapper h4');
```
or
```js
anchors.add(
  '#page-content-wrapper h2, ' + 
  '#page-content-wrapper h3, ' +
  '#page-content-wrapper h4'
);
```
would be the final result I am expecting. In either case, I would like to defer any change on this as possible as we can.

tnir: Before this heuristic improvement, I want to apply the non-heuristic way in #737, @olleolleolle. It looks like more time need for me to change this snippet.

PR #664, opened 2022-07-11
Remove jQuery dependency from search_arrows

tnir:

- See also #653
- cf. #654

It was planned that `search_arrows.js` is replaced without jQuery along with `search.js`, but it was found difficult as `search.js` has Popper.js v1 dependency.


deivid-rodriguez (approved):
Looks good! Small steps towards success!

PR #644, opened 2022-07-06
Replace bootstrap-sass 3 to bootstrap 4.6.1

tnir:
| before (1024-width on iPad Pro) | after (1024-width on iPad Pro) |
|-|-|
| bundler io_(iPad Pro) | bundler-site-tnir-boots-dje5be herokuapp com_(iPad Pro) |

Closes #643

### What was the end-user problem that led to this PR?

An end-user needs to load a legacy Bootstrap 3 just to browse the bundler docs.

### What was your diagnosis of the problem?

No maintainers/contributors maintained the frontend of the site for 6 years.

### What is your fix for the problem, implemented in this PR?

HAML and SCSS would be easier for everyone to maintain as Bootstrap 4 is more recent version than 3.

#### Screenshots for navbar/searchbox
comment-1177209898

#### New introduction
- https://fontawesome.com/v5/icons/search?s=solid (from Font Awesome 5 Free) (dropped Glyphicons icon font in Bootstrap 4)
- Opacity utilities from Bootstrap 5 https://getbootstrap.com/docs/5.1/utilities/colors/#opacity
  - which will be removed when Bootstrap upgraded from 4.6 to 5.1 (or upcoming 5.2)

##### Before this PR

https://user-images.githubusercontent.com/10229505/177769132-cdef09f2-f5c7-4a25-8cd9-67a4226dcfc8.mp4

##### After this PR
https://user-images.githubusercontent.com/10229505/177769152-e03cc14d-5681-40d6-b63b-ce1596ce312c.mp4

### Why did you choose this fix out of the possible options?

I cannot upgrade it from 3 to 5 (the latest major version as of creation of the PR) directly.

- Updates lots from #218

Signed-off-by: Takuya Noguchi [takninnovationresearch@gmail.com](https://github.com/sponsors/tnir)

PR #665, opened 2022-07-11
Remove jQuery dependency from search

tnir:

- See also #653
- cf. #654 and #664

It was found difficult as `search.js` has Popper.js v1 dependency.

- Blocked by #644
- Blocked by #681


tnir:
We can improve this after #644 merged.

tnir:
Thanks. Merging.

PR #651, opened 2022-07-11
Make guides (v1.16+) version-independent

tnir:
### What was the end-user problem that led to this PR?

Users can easily reach older guides from search engines.

Closes #628

### What was your diagnosis of the problem?

Almost version-agonistic guides are copied (see also diffs in #628)

### What is your fix for the problem, implemented in this PR?

```
$ git mv source/v2.3/guides source/guides
```
v2.3 guides are set up to the default guides.

```
$ git rm -r source/{v1.16,v1.17,v2.0,v2.1,v2.2}/guides
```
Remove v1.16 through v2.2 guides and all URLs are redirected to new URLs. And existing links are also updated.

For example,

- `https://bundler.io/v1.17/guides/git.html`
   - will be redirected to `https://bundler.io/guides/git.html`
- `https://bundler.io/v2.0/guides/bundler_2_upgrade.html`
   - will be redirected to `https://bundler.io/guides/bundler_2_upgrade.html`
- `https://bundler.io/pl/v2.2/guides/creating_gem.html`
   - will be redirected to `https://bundler.io/pl/guides/creating_gem.html`

```
$ git mv source/localizable/v2.3/guides source/localizable/guides
```
Follows up #625 for localizable guides. I found this should have done in #625 🙇 

### Why did you choose this fix out of the possible options?

Guides from v1.15 and v1.14 can be deferred to another PR.

Signed-off-by: Takuya Noguchi [takninnovationresearch@gmail.com](https://github.com/sponsors/tnir)

PR #666, opened 2022-07-11
Restore search for guides content

deivid-rodriguez:

The problem was moving around urls at #651 broke search for guides content.

My diagnosis was that after moving guides to not be version dependent, their contents are no longer indexed.

My fix is to add the new urls to the list of search resources.


tnir:
Thank you 🙇

PR #672, opened 2022-07-12
Enable all the default cops in RuboCop

tnir:
Only `EnforcedShorthandSyntax: never` is kept in `Style/HashSyntax` at this moment.

- Updates #617 and #649


deivid-rodriguez:
My personal approach to RuboCop is to use `DisableByDefault` and enable cops one by one as we are bothered by the inconsistencies or decide to enable specific cops for other reasons. Like we did in #617 and #347. I prefer that to blindly enabling all cops without consideration for each of them. What do you think?

tnir:
That could be an option, but does it tend to lengthen `.rubocop.yml`? RoboCup team has standardize rubocop configurations for a while. With this suggested change, I think we can start with long `todo.yml` and shorten step by step toward de facto standard in Ruby world.

Long `.rubocop.yml` does not look contributor-friendly to me. Contributors need to follow repo-specific cops by reading the `.rubocop.yml`.

I can start this with your way too.

deivid-rodriguez:
> Long .rubocop.yml does not look contributor-friendly to me. Contributors need to follow repo-specific cops by reading the .rubocop.yml.

If we follow all RuboCop rules, contributors still need to follow all RuboCop defaults, no? I think it's easier to enforce just the specific list of rules that we like, the bothers us the most, or whatever.

I think we can add `AllCops/DisplayStyleGuide: true` to our configuration to make it easier for contributors to fix issues.

PR #675, opened 2022-07-12
Use https over http in sitemap.xml

tnir:

While bunder.io fully supports HTTPS, `/sitemap.xml` has only `http`-based URLs.

We can simply replace them with HTTPS.

Use `https` over `http` as protocol.

No other choice.


deivid-rodriguez (approved):
Thanks!!

Issue #667, opened 2022-07-12
/manifest.json is 404


tnir:
`https://bundler.io/manifest.json` has been 404 since its inception (#225) for 6 years.

### Resources
- #225

PR #682, opened 2022-07-13
Put manifest.json in a correct place

tnir:

- See #667

Location of manifest.json was not correct.

Fixed the location of manifest.json.

n/a


olleolleolle (approved):
Looks like a fix!

PR #683, opened 2022-07-13
Remove imports-loader as it is now not required

tnir:

`imports-loader` is not required for `bootstrap@4` or did not look required for `bootstrap-sass`.

We need to follow up #562. Commit message in #562 explains:

> * `imports-loader` to inject the code to make jquery globals available into bootstrap-sass.

Just remove use of imports-loader and dependency of it.

While I am not sure if Bootstrap 3 (or bootstrap-sass) requires it, it is obvious there is no reference to imports-loader after #644.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,6 @@
       },
       "devDependencies": {
         "css-loader": "^5.2.0",
-        "imports-loader": "^4.0.0",
         "mini-css-extract-plugin": "^2.6.1",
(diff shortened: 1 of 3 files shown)


deivid-rodriguez (approved):
I am pretty sure it was actually needed for `bootstrap-sass`, yeah. But it doesn't matter whether it was needed or it was my mistake in the first place, since we're no longer using bootstrap-sass :)!

PR #671, opened 2022-07-12
Move more docs to guides

tnir:
### What was the end-user problem that led to this PR?

An end-user can get to the older version of guides, which are not dependent to versions.

- Closes #670

### What was your diagnosis of the problem?

Guide-ish documents have not been under `/guides` for 9 years (after #44). See #670 for diffs between the oldest and the current.

### What is your fix for the problem, implemented in this PR?

Removes files on `/v2.2` through `/v1.13`:

- `git ls-files source/v2.3 | grep -v -E "/bundle_|/docs|/index|/guides/|/whats" | sed -e "s#source/[^/]*/##;s#.html.*##" | sort | uniq | xargs -I {} git rm source/v{2.2,2.1,2.0,1.17,1.16,1.15,1.14,1.13}/{}.html.haml`

Removes files on `/v1.12`:

- `git ls-files source/v2.3 | grep -v -E "/bundle_|/docs|/index|/guides/|/whats" | sed -e "s#source/[^/]*/##;s#.html.*##" | sort | uniq | xargs -I {} git rm source/v1.12/{}.html.md`

Moves files on `/v2.3` to `/guides`:

- `git ls-files source/v2.3 | grep -v -E "/bundle_|/docs|/index|/guides/|/whats" | sed -e "s#source/[^/]*/##;s#.html.*##" | sort | uniq | xargs -I {} git mv source/v2.3/{}.html.haml source/guides/{}.html.haml`

Adds redirection `/:filename` to `/guides/:filename`

- see changes in `config.rb`

### Why did you choose this fix out of the possible options?

`bundler_workflow gemfile gemfile_ruby rationale rubygems rubymotion` are kind of guides, e.g.:

- As per conventions:
   - `http://bundler.io/gemfile.html`
      - should be `https://bundler.io/guides/gemfile.html`

Signed-off-by: Takuya Noguchi [takninnovationresearch@gmail.com](https://github.com/sponsors/tnir)

PR #689, opened 2022-07-15
Update static links after PR 671

tnir:

Some relative links are dead after #671.

Relative URLs should be updated.

Relative URLs were updated.

n/a 


tnir (commented):
With this change, there will be no dead link relevant to #671.

Issue #696, opened 2022-07-16
Replace fa-search icon with in-place SVG and update from FA Free 5.15.4 to FA Free 6.1.1


tnir:
#644 introduced (full) Font Awesome 5 free as a drop-in replacement for Glyphicons bundled in Bootstrap 3. Use of Font Awesome 5 for just a single icon https://fontawesome.com/v5/icons/search?s=solid

### Current code
https://github.com/rubygems/bundler-site/blob/678147f5a4aa502c366da286cccefd933e49ee17/source/layouts/_navbar.haml#L11

### Replacement code
```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512"><!--! Font Awesome Free 6.1.1 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free (Icons: CC BY 4.0, Fonts: SIL OFL 1.1, Code: MIT License) Copyright 2022 Fonticons, Inc. --><path d="M500.3 443.7l-119.7-119.7c27.22-40.41 40.65-90.9 33.46-144.7C401.8 87.79 326.8 13.32 235.2 1.723C99.01-15.51-15.51 99.01 1.724 235.2c11.6 91.64 86.08 166.7 177.6 178.9c53.8 7.189 104.3-6.236 144.7-33.46l119.7 119.7c15.62 15.62 40.95 15.62 56.57 0C515.9 484.7 515.9 459.3 500.3 443.7zM79.1 208c0-70.58 57.42-128 128-128s128 57.42 128 128c0 70.58-57.42 128-128 128S79.1 278.6 79.1 208z"/></svg>
```

- Source: https://github.com/FortAwesome/Font-Awesome/blob/6.x/js-packages/@fortawesome/fontawesome-free/svgs/solid/magnifying-glass.svg
- License: CC BY 4.0 Copyright 2022 Fonticons, Inc. (`Icons: CC BY 4.0, Fonts: SIL OFL 1.1, Code: MIT License`)

### Resources
- https://github.com/FortAwesome/Font-Awesome/issues/19191

tnir:
With this (#698) and #699, Performance in Lighthouse increased to 85/100 (Mobile) even from my local 🎉 :
![googlechrome github io_lighthouse_viewer_](https://user-images.githubusercontent.com/10229505/179366456-7d6af290-3680-4d16-8c52-3ac6a9f5e23d.png)

[...]

PR #699, opened 2022-07-16
Replace url-loader with Webpack's Asset Modules

tnir:

Web font has been bundled by `url-loader`, which has delayed first page rendering and would make web font delivery faster with CDN than GitHub Pages.

Webpack Asset Modules can be used to replace.

- `.otf` is now separated from `https://bundler.io/application.css` to an individual file as `https://bundler.io/font/e68a5a6780c1d57a2eed.otf`.
- As no woff, woff2, ttf, eot, and svg files are not bundled by Webpack, the configuration for these kind of files is also removed.
   - 1 `svg` file is linked from HAML, which has been simply copied on build by Middleman.

Updates part of #562

Webpack Asset Modules https://webpack.js.org/guides/asset-modules/ is considered as (relatively) modern.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,6 @@
         "sass": "^1.53.0",
         "sass-loader": "^13.0.2",
         "style-loader": "^3.3.1",
-        "url-loader": "^4.1.1",
         "webpack": "^5.73.0",
         "webpack-cli": "^4.10.0"
       }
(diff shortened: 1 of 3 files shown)


deivid-rodriguez (approved):
Sounds good, thanks!!

tnir:
Performance score report after this PR: https://github.com/rubygems/bundler-site/issues/696#issuecomment-1186252030

PR #700, opened 2022-07-16
Optimize cumulative layout shift (CLS) on top page

tnir:
https://web.dev/optimize-cls/

Cumulative Layout Shift (CLS) would be increased while it is currently 0.37.

### Before
pagespeed web dev_report_url=https%3A%2F%2Fbundler io%2F form_factor=desktop

### After
Will be pasted when this is deployed (in my test env it is 0).


tnir (commented):
It was confirmed that Cumulative Layout Shift became 0 in review apps. Merging to check it in production.

PR #701, opened 2022-07-16
Use font-display: swap for performance

tnir:

Downloading the font blocks text rendering.

`font-display: swap` can be used.

`font-display: swap` is applied. Fallback fonts in `font-family` were updated to be in serif, not sans-serif, which is the default of Bootstrap.

`swap` looks like the best value for it while `failback` is available: https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display

### DIFF
--- a/assets/stylesheets/_bootstrap-overrides.scss
+++ b/assets/stylesheets/_bootstrap-overrides.scss
@@ -24,7 +24,7 @@ $navbar-toggler-font-size: 1rem;
 // .btn does not need border
 $btn-border-width: 0;
 
-$font-family-sans-serif: "SancoaleSlab", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", "Liberation Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
+$font-family-sans-serif: "SancoaleSlab", Times, "Times New Roman", serif;
 
 // for contributors
 $card-border-width: 0;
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (commented):
I agree `font-display: swap` makes sense for us, but is the change in the font priority necessary? It feels weird to set a variable named `$font-family-sans-serif` to sherif fonts.

If I understand correctly setting `font-display` avoids blocking text rendering due to fonts not yet being available, what's the benefit of the other change?

tnir:
As "Sancoale Slab" font is a serif font, I think any of alternative font(s) should be a serif. The currently specified fonts are all sans serif, which are from Bootstrap and were introduced in #644, so I am removing all fonts added in #644 (which might be my mistake in it) in this PR.

> It feels weird to set a variable named $font-family-sans-serif to serif fonts.

I felt so, too, but I could not find a better way compatible with future Bootstrap versions. We might not have to take care about such a far future with Bootstrap 6 (hopefully in 2024 or 2025, for example 😀 ) etc. Or I should modify `$font-family-base` instead as per the following document:

> To switch the global `font-family`, update `$font-family-base` and recompile Bootstrap.
> -- https://getbootstrap.com/docs/5.1/content/reboot/#native-font-stack

> If I understand correctly setting font-display avoids blocking text rendering due to fonts not yet being available, what's the benefit of the other change?

`$font-family-sans-serif` (or `$font-family-base` mentioned above) is used for changing `font-family` for `body` (which is almost like `body { font-family: $font-family-sans-serif; }`.

deivid-rodriguez (approved):
Oh, I didn't realize Sancolae Slab was serif, nor did I notice you had just added these alternatives when migrating to Bootstrap 4.

This looks good as is to me, then!

PR #702, opened 2022-07-18
Replace middleman-search with lunr with Typesense DocSearch

### LINKED ISSUE #691: Replace lunr (middleman-search) with DocSearch
## Problems
As middleman-search has not been maintained (for example https://github.com/manastech/middleman-search/pull/38), we cannot upgrade lunr to the latest without our efforts (or other community's efforts than maintainers).

This also prevents from merging #661.

## Solutions
https://docsearch.algolia.com/ can be utilized.

Pros

- For open-source project, the cost is free.
- `lunr-index.json` is not required any more.
- UI components `search.js` and `search_arrows.js` can be replaced with Algolia's one, which seems sophisticated to me.
- We do not need to spend time for reinventing the wheel such as above.
- Fuzzy search by DocSearch is also nice.

Cons

- Proprietary (while no cost)

tnir:

As middleman-search ([RubyGems][RG], [GitHub][GH]) has not been maintained for more than 5.5 years (for example https://github.com/manastech/middleman-search/pull/38), bundler-site maintainers/contributors cannot upgrade lunr to the latest without their additional efforts (or other community's efforts than maintainers/contributors of this repo).

Closes #691

UI can be replaced by https://github.com/algolia/docsearch or its forks. Backend (current `/search/lunr-index.json`) can be replaced by Algolia DocSearch with automated crawling or Typesense cloud with manual crawling by us.

- Removes middleman-search gem and relevant gems (mini_racer)
- Removes JS/CSS codes and Middleman configuration for middleman-search
- Removes (the load of) Popover CSS from Bootstrap 5 along with the above JS code removal.
- Also the load of `/application.min.js` is moved to outside of `<head>`.

A crawler (which is expected to run once a day in production) is currently run by my local.

A cloud project on Typesense cloud called `ijmv1s078hbwfl9zp-1.a1.typesense.net` is owned by @tnir, which will be at no cost until 2022-08-16 and should be replaced by the project one (sponsored by RubyCentral?).

If Typesense offers us to some tiny instance at no cost and the permission to the UI, we might be able to use this change instead of Algolia's DocSearch.

[RG]: https://rubygems.org/gems/middleman-search
[GH]: [https://github.com/manastech/middleman-search]


olleolleolle:
So nice to get rid of the v8 too!

deivid-rodriguez:
Algolia alternative seems easier for now but I have to say I find this UI very nice looking.

deivid-rodriguez:
This looks fine performance wise, and the UI seems nice too. But the presence of a logo is a blocker and the majority of maintainers seem to prefer to take over middleman-search maintenance?

simi:
> This looks fine performance wise, and the UI seems nice too. But the presence of a logo is a blocker and the majority of maintainers seem to prefer to take over middleman-search maintenance?

indeed, let's focus on that one

tnir:
Features, look and feel are now all set. A small tweak in UI was submitted as a PR to the upstream, which is however NOT blocked to get this merged.

Any feedback is welcome again.

deivid-rodriguez:
@tnir You plan is unclear to me, since we don't have permission to go ahead without a logo. Everybody agrees to keep middleman-search for now.

tnir:
Just my previous comment was delayed. Just send back to draft at the moment while not sure this can be usable in the future.

deivid-rodriguez:
Ah, I see, makes sense now if comments are read in the right order :)

tnir:
By the way, another problem underlies in `assets/javascripts/search.js`. The hard way.

deivid-rodriguez:
I understand, but for now it seems our best option is to maintain these ~200 lines of JavaScript and this gem for searching...

PR #706, opened 2022-07-18
Replace middleman-search with lunr with Algolia DocSearch

### LINKED ISSUE #691: Replace lunr (middleman-search) with DocSearch

tnir:

As middleman-search ([RubyGems][RG], [GitHub][GH]) has not been maintained for more than 5.5 years (for example https://github.com/manastech/middleman-search/pull/38), bundler-site maintainers/contributors cannot upgrade lunr to the latest without their additional efforts (or other community's efforts than maintainers/contributors of this repo).

Closes #691

UI can be replaced by https://github.com/algolia/docsearch or its forks. Backend (current `/search/lunr-index.json`) can be replaced by Algolia DocSearch with automated crawling or Typesense cloud with manual crawling by us.

- Removes middleman-search gem and relevant gems (mini_racer and libv8-node).
- Removes JS/CSS codes and Middleman configuration for middleman-search.
- Removes HTML snippets thanks to new UI from DocSeach v3.
- Removes (the load of) Popover CSS from Bootstrap 5 along with the above JS code removal.
- Also the load of `/application.min.js` is moved to outside of `<head>`.

The crawler by Algolia is already running once a day in production.

Algolia's DocSearch for open source projects gets ready in a few days. If Typesense offers
- a some tiny instance to us at no cost
- and the permission to the UI
- DocSearch v3-ready UI for enduser,

we might be able to use #702, but this change looks clear at this moment.

[RG]: https://rubygems.org/gems/middleman-search
[GH]: [https://github.com/manastech/middleman-search]


shortcuts:
Hey, small feedback to improve your DocSearch results, you can update the Crawler config better scope the `lvl0` (e.g. `h1`), so you have more sections appearing in the modal (like on https://docsearch.algolia.com/)

tnir:
@shortcuts Thanks, but I could not see any good element for `lvl0` at this moment. I planned to update them by just adding section name after information architecting 😁 What do you think?

shortcuts:
> @shortcuts Thanks, but I could not see any good element for `lvl0` at this moment. I planned to update them by just adding section name after information architecting 😁 What do you think?

The main title of the page, the navbar section/active subsection is what we recommend

assets/javascripts/docsearch.js
```diff
@@ -0,0 +1,6 @@
+docsearch({
+  appId: "3JA5LRH987",
+  apiKey: "120ba00028e4e207674891463ae3fb17",
+  indexName: "bundler",
+  container: "#docsearch",
+});
```
deivid-rodriguez, line 6: Should these keys be private?

source/layouts/base.haml
```diff
@@ -30,3 +30,6 @@
 
     .footer
       = partial 'layouts/footer'
+    = javascript_include_tag "https://cdn.jsdelivr.net/npm/@docsearch/js@3.1.1/dist/umd/index.min.js"
+    = javascript_include_tag "https://cdn.jsdelivr.net/npm/bootstrap@5.2.0-beta1/dist/js/bootstrap.min.js"
```
deivid-rodriguez, line 34: Why is this? Is it related to not longer needing Popover?

deivid-rodriguez (commented):
Looks nice!

I agree that Algolia Docsearch seems easier for the moment.

Is it possible to recrawl the website on every deploy instead of daily?

While testing I noticed some extra search results which should not be needed. For example, when looking for "docker", one of the results is a link to the sidebar: https://bundler.io/guides/bundler_docker_guide.html#sidebar-wrapper. Can we avoid generating these redundant results?

assets/javascripts/docsearch.js
tnir: This (`apiKey`) is a search-only (read only) API key commonly used for front-end.

> This is the public API key which can be safely used in your frontend code.This key is usable for search queries and it's also able to list the indices you've got access to.

deivid-rodriguez: Yes, I had just noticed it in Algolia's UI 👍.

simi:
Is the search branded right now? That would be big :-1: for me.

source/layouts/base.haml
tnir: I used to think this is not needed. By removing the final load `import { Popover } from 'bootstrap';` from `assets/javascripts/search.js`, it looks like bootstrap esm JS will not loaded any more, which will break navbar collapse. So with the current situation of slowness of asset (`https://bundler.io/application.min.js`) downloading from GitHub Pages.

The below image shows how fast JavaScripts are delivered from CDN (DocSearch JS, Bootstgrap JS, and our own JS from top to bottom):
algolia

See also https://getbootstrap.com/docs/5.1/getting-started/webpack/#importing-javascript

tnir:
@simi Yes I get you, but Typesense will cause the same result...

tnir:
@simi As the next step to this PR, we can prepare our own cluster on Algolia (or Typesense cloud). We can estimate how much it will be if some sponsors pay for it.

> If this is not possible for you, you're free to [open your own Algolia account](https://www.algolia.com/pricing) and run [DocSearch on your own](https://docsearch.algolia.com/docs/legacy/run-your-own) without this limitation. In that case, though, depending on the size of your documentation, you might need a paid account (free accounts can hold as much as 10k records).

https://docsearch.algolia.com/docs/DocSearch-program

assets/javascripts/docsearch.js
tnir: Okay, so resolving this conversation.

deivid-rodriguez:
In any case, it seems best for us to be in full control of when we run the crawler (after every deploy), so running our own cluster would seem like the way to go.

shortcuts:
> Is it possible to recrawl the website on every deploy instead of daily?

Sure, we have a GitHub action available: https://github.com/algolia/algoliasearch-crawler-github-actions

tnir:
Updated the primary color for box-shadow etc. in searchbox and modal as I have forgotten the adaption of Bundler branding color.

tnir:
Admin role is now requested to Org members/Repo admins in #730 🙏

source/layouts/base.haml
tnir, line 34: You should use bootstrap 5.1.3 to align with CSS and to avoid a regression in 5.2.0.

tnir (commented):
I will fix it soon.

source/layouts/base.haml
tnir: Fixed in 3f2bf0e.

tnir:
@deivid-rodriguez How can I trigger a review app for this (old) PR (just development purpose)?

deivid-rodriguez:
I do it manually from Heroku UI, I created it for you.

hsbt:
We will migrate bundler.io to rubygems.org. We should consider this at https://github.com/rubygems/guides/

Issue #707, opened 2022-07-19
Activate dead developer guides


tnir:
These pages were introduced in #317, but (almost?) all of them have been orphan while they are still on the sitemap (`/sitemap.xml`).

- https://bundler.io/doc/contributing/bug_triage.html
- https://bundler.io/doc/contributing/community.html
- https://bundler.io/doc/contributing/getting_help.html
- https://bundler.io/doc/contributing/how_you_can_help.html
- https://bundler.io/doc/contributing/issues.html
- https://bundler.io/doc/contributing/readme.html
- https://bundler.io/doc/development/new_features.html
- https://bundler.io/doc/development/pull_requests.html
- https://bundler.io/doc/development/readme.html
- https://bundler.io/doc/development/setup.html
- https://bundler.io/doc/documentation/readme.html
- https://bundler.io/doc/documentation/vision.html
- https://bundler.io/doc/documentation/writing.html
- https://bundler.io/doc/playbooks/releasing.html
- https://bundler.io/doc/playbooks/team_changes.html
- https://bundler.io/doc/policies.html
- https://bundler.io/doc/readme.html
- https://bundler.io/doc/troubleshooting.html

## Solutions
`https://twitter.com/rappmasterflash/status/851494907906670592` does not look available any more, but we can respect #317. Due to the repository migration to `rubygems/rubygems`, this kind of static URLs would be helpful for new developers.

Note: only `https://bundler.io/doc/troubleshooting.html` is a kind of user guides than developer guides.

tnir:
With #708, https://bundler.io/doc/readme.html is now linked rom https://bundler.io/docs.html 🎉 So closing.

tnir:
As a result of #708, the crawled URLs increased from 150 to 171 and indecies increased from 974 to 1.12k:

![crawler algolia com_admin_crawlers_7f4b6579-bba3-4c1f-a0ff-462d9f408281_monitoring_summary](https://user-images.githubusercontent.com/10229505/179749024-b4da7fa7-9b00-41d1-8d9e-992e2c1679ab.png)

PR #708, opened 2022-07-19
Add links to dev guides from user guides

tnir:

- #707

Also relevant to

- #691

Links can be added at minimum. All developer guides, generated from the main repo, already exist on Web site.

Adds a link from the sidebar in each guide (not command references) to [the top of contribution and developer guide](https://bundler.io/doc/readme.html).

1. A crawler and/or a user can find the entrance of from the Bundler Web site with this minimal change.
2. We can add dev guides on `https://bundler.io/docs.html`, but I skipped it **at this moment** since they are versioned.

### DIFF
--- a/source/v2.3/docs.html.haml
+++ b/source/v2.3/docs.html.haml
@@ -16,6 +16,7 @@
       %ul.ul-padding
         - guides.each do |guide|
           %li= link_to_guide guide
+        %li= link_to "Contributing to Bundler", "/doc/readme.html"
 
     .col-md-4.offset-md-0.col-9.offset-2
       %h3


deivid-rodriguez (commented):
I think the distinction between "Guides" and "Developer Guides" is not very clear, since many Bundler users will call themselves developers.

Should we instead a new link to the main list, called "Contributing to Bundler" or "Bundler contributor guidelines" or something like that?

tnir:
I agree with you that  the distinction between "Guides" and "Developer Guides" is not very clear!

Let me revert all in the previous commit and just add a link from the latest docs to "Contribution guides" i.e. `https://bundler.io/docs.html`.

PR #709, opened 2022-07-19
Add tnir to Website team

tnir:

Members working on Website are different from the page.

Readers can get confused.

Add tnir to Website team

Updating them with active members is also possible.


olleolleolle (approved):
Yay!

deivid-rodriguez (commented):
Looks good! Maybe your entry should be the first one of the three since you are the main website maintainer now.

tnir:
Updated as you suggested, @deivid-rodriguez. Merging.

deivid-rodriguez:
Nice, thank you!

PR #596, opened 2022-01-25
Bundler 2.3 docs and Bundler 2.3 blog post draft

deivid-rodriguez:
### What was the end-user problem that led to this PR?

We haven't published Bundler 2.3 docs, nor a release blog post.

### What is your fix for the problem, implemented in this PR?

Publish docs and write a blog post draft.

PR #720, opened 2022-07-21
Copyedit Bundler v2.3 release blog article

tnir:

https://bundler.io/blog/2022/01/23/bundler-v2-3.html has a `h1` heading at the end of the article, which looks weird as other headings are all `h2` even if you want to emphasize "What's coming next?"

This would degrade search experience if we introduced a more semantic-based crawler in the future.

That section can be `h2` to align with other sections in the article.

Align that heading with other headings by changing from `h1` to `h2`.

Any reason may exist behind #596


deivid-rodriguez (approved):
👍

PR #737, opened 2022-07-23
Introduce prettier as JS linter

tnir:
## Contributor problem

JavaScript files are not so consistent.

Any linter for JavaScript files is not setup yet.

Introduce prettier as a linter for JavaScript files.

`singleQuote` looks preferred unlike [Prettier default](https://prettier.io/docs/en/options.html#quotes), so `singleQuote` is set `true` in this PR at this moment.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,3 +32,15 @@ jobs:
         run: bundle exec middleman build
         env:
           NO_CONTRACTS: "true"
+
+  prettier:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: npm
+      - run: npm install
+      - name: Prettier JS
+        run: npx prettier --check assets/javascripts
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,1 @@
+singleQuote: true
--- a/assets/javascripts/application.js
+++ b/assets/javascripts/application.js
@@ -1,9 +1,9 @@
-import './search'
-import './search_arrows'
+import './search';
+import './search_arrows';
 import AnchorJS from 'anchor-js';
 
 const anchors = new AnchorJS();
 
 anchors.options = {
-  visible: 'touch'
+  visible: 'touch',
 };
--- a/assets/javascripts/commands_layout.js
+++ b/assets/javascripts/commands_layout.js
@@ -7,5 +7,7 @@ el.addEventListener('input', function (e) {
   document.location.href = e.target.value;
 });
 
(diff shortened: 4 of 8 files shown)


.github/workflows/build.yml
```diff
@@ -32,3 +32,15 @@ jobs:
...
           NO_CONTRACTS: "true"
+
+  prettier:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+      - name: Install prettier 2.7.1
```
deivid-rodriguez, line 43: I guess "Install prettier" so you only need to update when place when you upgrade it?

deivid-rodriguez (commented):
Sounds good!

.github/workflows/build.yml
```diff
@@ -32,3 +32,15 @@ jobs:
...
+        run: npm install prettier@2.7.1
```
simi, line 44: Any reason to not make it part of devDeps in `package.json` actually and track version in there?

deivid-rodriguez: Good idea, indeed, and will probably work better if we want dependabot to send us updates in the future!

.github/workflows/build.yml
tnir: Makes sense.

.github/workflows/build.yml
tnir: Makes sense.

tnir:
@deivid-rodriguez @simi Thank you for your feedback.

simi:
> @deivid-rodriguez @simi Thank you for your feedback.

It was a pleasure. Btw. we can add this commit to `.git-blame-ignore-revs` as well. But that is totally optional in here.

tnir:
I do not think it should do so as changes in `assets/javascripts/*.js` in b94c55b197e3be0416cfb9aefc44a31b64d58f4c is only in 81 lines (and .github/workflows was slightly improved).
```
$ git show b94c55b197e3be0416cfb9aefc44a31b64d58f4c --numstat
commit b94c55b197e3be0416cfb9aefc44a31b64d58f4c
Author: Takuya Noguchi <takninnovationresearch@gmail.com>
Date:   Sat Jul 23 15:13:01 2022 +0000

    Add prettier as JS linter
    
12      0       .github/workflows/build.yml
1       0       .prettierrc
3       3       assets/javascripts/application.js
4       2       assets/javascripts/commands_layout.js
39      33      assets/javascripts/search.js
35      26      assets/javascripts/search_arrows.js
22      0       package-lock.json
4       2       package.json
```
Thanks.

deivid-rodriguez:
Yeah, I think using `.git-blame-ignore-revs` should only be used for huge commits.

PR #739, opened 2022-07-24
Remove status checks from Gitpod prebuild

tnir:
Removes one of status checks from Gitpod prebuild as it takes a longer time than others and Gitpod prebuild is not mandatory at this moment.

Updates #648

- #648
- https://www.gitpod.io/docs/references/gitpod-yml#prebuildsaddcheck


deivid-rodriguez (approved):

tnir:
With this, the checks would be green faster than before.

PR #740, opened 2022-07-24
Pin ruby-version to 3.1.2 only without matrix

tnir:
This Web site is not a library and is expected to run with only a single Ruby version specified in `/.ruby-version`. The matrix has been introduced in #518 for a different purpose from the expectation. However, the original purpose was resolved in #558, but #558 missed this removal.

This will help us see GitHub status checks in pull requests since it becomes shorter.

Follows up #558

Updates #518

### before

before-740

### after
(The section `prettier` is not relevant)

after-740

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,16 +4,13 @@ on: [pull_request]
 jobs:
   build:
     runs-on: ubuntu-20.04
-    strategy:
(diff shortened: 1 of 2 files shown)


olleolleolle (approved):
Great fix!

tnir:
Thank you.

Issue #761, opened 2022-07-26
Add warning messages in deprecated minor versions


tnir:
Add warning messages in deprecated minor versions (v1.x and v2.0) to both guides and man pages as per https://github.com/rubygems/rubygems/issues/5647.

- cf. #760

PR #762, opened 2022-07-26
Append maintenance status per version

### LINKED ISSUE #760: State the status per minor versions
We can specify more status for each minor versions as per https://github.com/rubygems/rubygems/issues/5647.

## Current
---

bundler io_v2 3_man_bundle-viz 1 html

---

version-selector

## Solution
Add the version status in a parenthesis like npm CLI:

npm-cli-v8

https://docs.npmjs.com/cli/v8

tnir:

A visitor cannot see what version is active while 10 minor versions are listed.

Closes #760

Looks good to start with indicating the maintenance status per version rubygems/rubygems#5647.

Append the maintenance status per version next to each version in the version selector in the versioned man pages.

| before | after |
|-|-|
| bundler io_v2 3_man_bundle-viz 1 html | v2-closed |
| version-selector | v2-open |

<details>
<summary>Previous version of screenshots after the change (only for reference)</summary>

bundler-site-tnir-versi-phrzrh herokuapp com_v2 3_man_bundle-install 1 html
version-selector-after
</details>

No other option can be imagine...

### DIFF
--- a/helpers/config_helper.rb
+++ b/helpers/config_helper.rb
@@ -6,4 +6,20 @@ module ConfigHelper
   def versions
     config[:versions]
   end
+
+  def versions_grouped_by_status
+    versions.reverse.group_by { status(_1) }
+  end
+
+  private
+
+  def status(version)
+    if version == "v2.3"
+      "Current release"
+    elsif %w[v2.2 v2.1].include?(version)
+      "Legacy release"
+    else
+      "Deprecated release"
+    end
+  end
 end
--- a/source/partials/_commands_sidebar.haml
+++ b/source/partials/_commands_sidebar.haml
@@ -2,10 +2,12 @@
 
 %h4 Choose version
 %select.version-selects.form-select.mb-3
-  - versions.reverse.each do |version|
-    %option{selected: version == current_visible_version,
-                      value: documentation_path(current_page_without_version, version) || documentation_path('bundle-install.1', version)}
-      = version
+  - versions_grouped_by_status.each do |status, _versions|
+    %optgroup{label: status}
+      - _versions.each do |version|
+        %option{selected: version == current_visible_version,
+                          value: documentation_path(current_page_without_version, version) || documentation_path('bundle-install.1', version)}
+          = version
 %h4 Primary Commands
 %ul
   - primary_commands.select{ |page| path_exist?(page, current_visible_version) }.each do |page|


olleolleolle:
One possible extension would be to use HTML optgroup elements to group the "maintenance levels".

lib/versions.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+VERSIONS = Dir.chdir("source") { Dir.glob("v*").sort_by{|x| Gem::Version.new(x[1..-1]) } }.freeze
```
olleolleolle, line 3: Opportunity to use _1 in the block.

tnir: 👍, but I do not have a plan to change the line but just copy whole the line from `lib/tasks/versions.rake` to `lib/versions.rb`.

helpers/config_helper.rb
```diff
@@ -6,4 +8,18 @@ def current_version
...
\ No newline at end of file
+
+  def version_with_status(version)
+    "#{version} (#{status(version)})"
+  end
+
+  def status(version)
+    if version == versions.last
+      "Current release"
+    elsif versions.last(3).include?(version)
```
deivid-rodriguez, line 19: I'm not sure this is a rule that will always hold, maybe it's better to hardcode the current situation.

deivid-rodriguez (approved):
I think this is a good idea.

helpers/config_helper.rb
tnir: It makes sense to me, too. Fixed it in the latest commit.

deivid-rodriguez:
Ok with me, not sure if you plan to implement @olleolleolle's suggestion about html optgroup elements?

lib/versions.rb
tnir: Due to "hardcoding", we do not need change in `VERSIONS` definition any more. For an improvement task in the future, just filed in #769.

tnir:
@deivid-rodriguez Thanks.
@olleolleolle Which in the below CodePen do you expect or what else? Option 1 can hardly reduce feeling like too packed, but Option 2 cannot indicate the status of the version a user is seeing to them itself.
https://codepen.io/tnir/pen/BarwNOm

I am slightly inclined to choose Option 2 if we will go with a more notice in the main area (or in the fixed header) by #761.

olleolleolle:
To me, option 2 seems easier to understand.

tnir:
Thanks @olleolleolle for the preference! Happy to merge this and to improve the presentation in the main area in #761.

Issue #775, opened 2022-07-27
Enable Style/StringLiterals cop in RuboCop


tnir:
Quotations in Ruby files can be controlled by RuboCop.

> Should we introduce [haml-lint](https://github.com/sds/haml-lint) and split using double quotes and enforcing its usage from now on to a separate PR?

_Originally posted by @deivid-rodriguez in https://github.com/rubygems/bundler-site/pull/773#pullrequestreview-1052108428_

cf.

- #787

tnir:
cc @deivid-rodriguez

PR #776, opened 2022-07-27
Introduce HAML-Lint for HAML files

### LINKED ISSUE #774: Introduce HAML-Lint
Our HAML files vary in formatting.

- cf. #775

tnir:
- Introduce HAML-Lint for HAML files
- Reorganize Action workflow files

Closes #774

Note: this does not include fix in single/double quotes due to missing RuboCop configurations (see #775)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,22 +22,7 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      - name: Run RuboCop
-        run: bundle exec rubocop
-
       - name: Build
         run: bundle exec middleman build
         env:
           NO_CONTRACTS: "true"
-
-  prettier:
-    runs-on: ubuntu-22.04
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
Shouldn't this fail because of the inconsistent quotes when the idea of adding `haml-lint` came up?

deivid-rodriguez (approved):
Oh, I understand. We need to enable the `StringLiterals` cop in RuboCop since `haml-lint` will simply respect your RuboCop configuration.

Then ok, regarding the rest of the rules I don't have a strong opinion. I guess we're already enforcing the rules you set, so it seems like a good starting point.

tnir:
Yes, this is what you wanted to do in #672 (I mean if we accepted #672, we could have done this at once. We just need more steps, which I prefer too). Merging this to move this big step forward! 🚄

PR #778, opened 2022-07-27
Heroku domain is now on herokuapp.com

tnir:
Heroku's subdomain has been under `herokuapp.com` instead of `heroku.com` for so many years.


tnir:
@hone does not look in the team any more...

PR #782, opened 2022-07-27
Enable RuboCop entirely in HAML-Lint

tnir:
Prior to #775, remove `exclude` from the `RuboCop` rule in HAML-Lint.


deivid-rodriguez (approved):
Nice! Same Ruby style everywhere.

Issue #233, opened 2016-08-02
Clear Fastly cache on deploy


indirect:
Turns out Fastly caches old pages for up to 10 minutes after we deploy changes 😲 We should make an API call to clear the cache after our deploy is done.

kruczjak:
Probably this call will do that :)
https://docs.fastly.com/api/purge#purge_bee5ed1a0cfd541e8b9f970a44718546

PR #786, opened 2022-07-28
Remove CDN purge command from deploy job

### LINKED ISSUE #744: Fastly does not look used
- Split deploy and purge
  - [ ] #743
- Remove purge Rake task

### Resources
https://github.com/rubygems/bundler-site/runs/7487766972?check_suite_focus=true

tnir:
Closes #744

Fixes up #241

cf.
- #233
- #339
- #558

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -31,6 +31,3 @@ jobs:
           encrypted_key: ${{ secrets.encrypted_key }}
           encrypted_iv: ${{ secrets.encrypted_iv }}
           COMMIT_AUTHOR_EMAIL: team@bundler.io
-
-      - name: Purge CDN
-        run: bundle exec rake ci:clean_fastly_cache


deivid-rodriguez (commented):
This makes sense to me. Fastly encrypted ENV variables were removed from TravisCI config at https://github.com/rubygems/bundler-site/pull/339, because we started setting them at the web UI. When I migrated from TravisCI to Github Actions, I forgot to check Web UI variables and didn't migrate them, and @tnir confirmed that we're not using them these days. Yet no issues about this so I guess it's not necessary. I don't get why the task is not failing though.

All in all, I don't think Fastly is involved at all here, but I'd like @indirect to confirm if possible.

indirect:
Yep, this is all from back when we used Fastly to proxy bundler.io to bundler.github.io. Nowadays GitHub Pages hosts bundler.io directly. 👍🏻

deivid-rodriguez:
Aaah, I see. Thanks @indirect!

PR #789, opened 2022-07-29
Use the same repository but the different branch for GitHub Pages for production

tnir:
Closes #591

Changes the publishing source for GitHub Pages for `https://bundler.io`.

- before: https://github.com/rubygems/bundler.github.io/tree/master
- after: https://github.com/rubygems/bundler-site/tree/gh-pages

## Notes
- `ci:update_ssh_site` Rake task does now use HTTPS over SSH per [GitHub recommendation](https://docs.github.com/en/get-started/getting-started-with-git/about-remote-repositories) and thanks to pre-defined token in GitHub Actions. Unify it with `update_site` Rake task in another PR.
- `configure_ssh_deploy_key` remains. Remove it in another PR.
- Permissions in GitHub Actions are not changed. Improve it in another PR.
- New Custom GitHub Actions Workflows (beta) for GitHub Pages are not included. See https://github.com/middleman/middleman/issues/2561
   - or https://github.com/rubygems/bundler-site/issues/790

Signed-off-by: Takuya Noguchi [takninnovationresearch@gmail.com](https://github.com/sponsors/tnir)

Issue #591, opened 2021-12-11
Merge this repository with https://github.com/rubygems/bundler.github.io


deivid-rodriguez:
This is not a big deal, really, but it feels a bit cumbersome to use two different repositories.

As far as I understand, we could either:

* Use a branch in this repository to deploy to GitHub Pages, instead of a separate repository.

or

* Stop deploying to GitHub Pages and use Heroku instead.

The former seems simpler, but the latter seems slightly more desirable to me, because we currently have a permanently red Heroku staging pipeline, just to keep Heroku review apps working, which is very useful. I guess we could fix that actually having a staging site that uses heroku, but at that point we may as well switch everything to Heroku.

Anways, this is not a big deal, but it could be an interesting/fun exercise for someone willing to play around with Heroku and Github Pages.

## Summary

- Code changes: https://github.com/rubygems/bundler-site/compare/fe5036b...b16c5ef

### `rubygems/bundler-site`
- [x] Set @tnir as Maintain or higher (= Admin) in `rubygems/bundler-site` (#730)
- [x] Manually push `master` branch in https://github.com/rubygems/bundler.github.io to this repo as `gh-pages` and keep up to date (as possible).
- [x] Enable GitHub Pages with publishing source of `gh-pages` branch and `/` (root)
- [-] Set @tnir as Admin in `rubygems/bundler-site` (#730)
- [x] Set a custom domain of `bundler.io` in GitHub Pages at https://github.com/rubygems/bundler-site/settings/pages
- [x] Create a PR to change the deploy target from `rubygems/bundler.github.io` to `rubygems/bundler-site` in [ci/deploy.rake](https://github.com/rubygems/bundler-site/blob/bbb543be924b8a812aba486bf7cde1f5e03e8c2e/lib/tasks/ci/deploy.rake#L21-L41) (#789)
- [x] Review and merge the PR above #789

### `rubygems/bundler.github.io`
- [x] Set @tnir as Admin in `rubygems/bundler.github.io` (#730)
- [x] Unset a custom domain in the old repo.
- [x] Remove CNAME in rubygems/bundler.github.io: https://github.com/rubygems/bundler.github.io/commit/cf6cdd77cc51c9104feecb947cc9af27ce0f6248
- [x] Disable GitHub Pages by selecting `None` as the publishing source at https://github.com/rubygems/bundler.github.io/settings/pages
- [x] (optional) Add a commit with the deprecation notice on the default branch to README.
- [x] (optional) Archive the repo

[...]

Issue #790, opened 2022-07-29
Use Custom GitHub Actions Workflows (beta) for GitHub Pages


tnir:
With [the release of Custom GitHub Actions Workflows (beta) for GitHub Pages](https://github.blog/changelog/2022-07-27-github-pages-custom-github-actions-workflows-beta/) on July 27, 2022, we can benefit from it to make the publishing source from `gh-pages` branch to `main` especially after #789.

hsbt:
I prefer to switch this. 

FYI: The current github pages already support to deploy same branch with root or any directory. We can choose this change before beta feature.

indirect:
@hsbt I think we need to use GitHub Actions to run `middleman build`. So just choosing "deploy gh-pages branch to main branch" is not enough, and we need either a "normal" GitHub action or the beta for Pages.

tnir:
Yes, https://github.com/actions/deploy-pages exists for a while, which can be used for your own Actions for Pages, and my understanding is @indirect' comment above. Custom GitHub Actions workflows (beta) for GitHub Pages (as of https://github.com/actions/deploy-pages/pull/42) does not include the Middleman workflow, so we need to write some code.

tnir:
Even including `CNAME` file did not help me use the custom domain (#796) as follows, and I needed to restore the publishing source from `master` to `gh-pages` even after #795/#796.

- Workflow from gh-pages branch got `"page_url":"https://bundler.io/"`: https://github.com/rubygems/bundler-site/runs/7573442939?check_suite_focus=true
- but one from master branch got `"page_url":"https://blog.rubygems.org/bundler-site/"`: https://github.com/rubygems/bundler-site/runs/7573407610?check_suite_focus=true

@hsbt Can you check what happens and if CNAME is preserved as `bundler.io` when you change the source from `gh-pages` to `master` at https://github.com/rubygems/bundler-site/settings/pages?

hsbt:
Custom domain configuration may depends on [this](https://github.com/rubygems/bundler-site/pull/795#issuecomment-1198918726).

tnir:
Thanks. While it seems https://github.com/rubygems/bundler-site/settings/pages shows incorrect information to "Maintain" members, it will be no problem unless a "Maintain" member updates it. Closing.

PR #791, opened 2022-07-29
Use SSH protocol for git push on deploy job

tnir:
Fixes up #789

While in #789, we changed the protocol from SSH to HTTPS along with changes on the repository (#591), it did not work. This PR reverts the protocol to SSH.

### DIFF
--- a/lib/tasks/ci/ssh_repo.rake
+++ b/lib/tasks/ci/ssh_repo.rake
@@ -1,5 +1,5 @@
 directory "vendor/ssh_bundler.github.io" => ["vendor"] do
-  system "git clone https://github.com/rubygems/bundler-site.git vendor/ssh_bundler.github.io"
+  system "git clone git@github.com:rubygems/bundler-site.git vendor/ssh_bundler.github.io"
 end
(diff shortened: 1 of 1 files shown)


hsbt (approved):

tnir:
I am not sure the SSH key used for `rubygems/bundler.github.io` can be used for this repo... @hsbt What do you think?

indirect:
I think we used to use an SSH key that belongs to the @bundlerbot GitHub account. Nowadays I think we could generate an SSH key, put the public key under "deploy keys" for the repo, and store the private key in GitHub Action secrets? Or maybe we could just use this GitHub action to do the deploy: https://github.com/marketplace/actions/github-pages-action.

tnir:
Thanks @indirect. While I could not see public SSH keys at https://github.com/bundlerbot.keys, so do you mean deploy keys with Write permission against (at least or only) `rubygems/bundler.github.io` repo, don't you? Anyway, it looks like we could safely merge this. To modernize deploy workflow, I just prepared #790.

PR #795, opened 2022-07-29
Migrate the deploy Action from homemade to the official

tnir:
With [the release of Custom GitHub Actions Workflows (beta) for GitHub Pages](https://github.blog/changelog/2022-07-27-github-pages-custom-github-actions-workflows-beta/) on July 27, 2022, we can benefit from it to make the publishing source from `gh-pages` branch to `main` after #591 (= #794 + #789).

Closes #790

Signed-off-by: Takuya Noguchi [takninnovationresearch@gmail.com](https://github.com/sponsors/tnir)

PR #804, opened 2022-07-29
Remove deprecated Rake tasks for legacy CI

tnir:
Follows up #795


tnir (commented):
No regression confirmed.

PR #817, opened 2022-07-30
Remove touch option from AnchorJS

### LINKED ISSUE #816: Anchor icon is now always visible
After #809 with #632, Anchor icon by AnchorJS is now always visible.

- Issue report in the upstream: https://github.com/bryanbraun/anchorjs/issues/168

We should now use `visible: hover` (default) instead.

tnir:
This option works no more in many browsers as expected in #218.

- https://caniuse.com/touch
- https://github.com/bryanbraun/anchorjs/issues/168

Closes #816

Signed-off-by: Takuya Noguchi [takninnovationresearch@gmail.com](https://caniuse.com/touch)

### DIFF
--- a/assets/javascripts/two_column_layout.js
+++ b/assets/javascripts/two_column_layout.js
@@ -2,10 +2,6 @@ import AnchorJS from 'anchor-js';
 
 const anchors = new AnchorJS();
 
-anchors.options = {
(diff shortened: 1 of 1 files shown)


tnir (commented):
This reverts the behavior as before #809.

PR #818, opened 2022-07-30
Remove ci:fastly_cache_purge_uri task

tnir:
With #786, we do not need  ci:fastly_cache_purge_uri task any more.


tnir (commented):
Deleting. If you need one like this, per https://developer.fastly.com/reference/api/purging/, https://rubygems.org/gems/fastly/versions/4.0.0.alpha4 can be used, but `curl`-based one-liner looks the best to me.

PR #815, opened 2022-07-30
Use path in actions/upload-pages-artifact

tnir:
Partially reverts #795

Closes #814

## Checklist
- [x] Review app renders correctly (not 404)
- [ ] Deploy to production

Signed-off-by: Takuya Noguchi [takninnovationresearch@gmail.com](https://github.com/sponsors/tnir)

PR #822, opened 2022-07-31
Support non-default build_dir in a Rack server

tnir:
Make deployment to Heroku review apps robust to support non-default `build_dir` configuration in Middleman.

Follows up #815 


tnir (commented):
Review apps for this change look good.

PR #823, opened 2022-07-31
Use h1 instead of h2 for command name in command reference

tnir:
Use `h1` tag for the command name in command reference, instead of `h2`, which is generated by `ronn`.

## before
bundler io_v2 3_man_bundle-clean 1 html

## after
4567-rubygems-bundlersite-fd5h6ytcwan ws-us58 gitpod io_v2 3_man_bundle-clean 1 html


olleolleolle (approved):
An improvement to the structure of the document!

deivid-rodriguez (approved):
💪

tnir:
Yes, I would like to get rid of `ronn`, which has been maintained for more than a decade... to avoid such a hack...

deivid-rodriguez:
Indeed. The most successful fork is https://github.com/apjanke/ronn-ng, I considered migrating a while ago, but it still needed some work to be a drop in replacement.

PR #824, opened 2022-07-31
Update gitpod workspace with the default

tnir:
Updates #648

As `gitpod-io/workspace-full` image, the default image from gitpod-io/workspace-images is now based on both Ruby 3.1.2 and Node.js 16.x, we do not need to built it by ourselves.

- https://github.com/gitpod-io/workspace-images/pull/889

## Check-list

- [x] Open a new workspace from `https://gitpod.io/#github.com/rubygems/bundler-site/tree/tnir-gitpod` to see the server launches without any special operation
- [x] Stop the workspace opened above and restart it to see the server launches normally without any special operation

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,4 +1,0 @@
-FROM gitpod/workspace-ruby-3.1
-
-RUN curl -fsSL https://deb.nodesource.com/setup_16.x | sudo -E bash - && \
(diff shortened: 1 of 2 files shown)


tnir:
@deivid-rodriguez What do you think?

deivid-rodriguez (approved):

PR #841, opened 2022-08-01
Replace missing forked ronn-ng with public nronn

### LINKED ISSUE #1003: Failed to pull deivid-rodriguez/ronn-ng@fix-charset
https://github.com/apjanke/ronn-ng/pull/83

```
$ bundle
Fetching https://github.com/deivid-rodriguez/ronn-ng.git

Git error: command `git fetch --force --quiet --no-tags --depth 1 -- https://github.com/deivid-rodriguez/ronn-ng.git refs/heads/fix-charset:refs/heads/fix-charset` in
directory /workspace/bundler-site/vendor/bundle/ruby/3.2.0/cache/bundler/git/ronn-ng-74fdccba57aa79e68ec2317a909ea1ae39691ed0 has failed.
Revision fix-charset does not exist in the repository https://github.com/deivid-rodriguez/ronn-ng.git. Maybe you misspelled it?
```

tnir:
- Closes #1003

~~Use nronn to unify use of rdiscount with kramdown, and to unify use of hpricot with nokogiri.~~

1. rdiscount and kramdown both provide similar functionality: avoid to use rdicount
2. hpricot and nokogiri both provide similar functionality: avoid to use hpricot
3. Tested with Ruby 3.1 in CI (ronn has never continuously tested 😭).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,15 +1,4 @@
 GIT
-  remote: https://github.com/deivid-rodriguez/ronn-ng.git
-  revision: f1c1c79414c85351720e1d9ff6cd1737fe337295
-  branch: fix-charset
-  specs:
-    ronn-ng (0.10.1.pre1)
-      kramdown (~> 2.1)
-      kramdown-parser-gfm (~> 1.0.1)
-      mustache (~> 1.0)
-      nokogiri (~> 1.11, >= 1.11.0)
-
-GIT
   remote: https://github.com/middleman/middleman.git
   revision: 038f4f606ff6cfc75cbe5c8690c5eae2e34f78f3
   ref: 038f4f606ff6cfc75cbe5c8690c5eae2e34f78f3
@@ -132,6 +121,11 @@ GEM
     nokogiri (1.13.10)
       mini_portile2 (~> 2.8.0)
       racc (~> 1.4)
+    nronn (0.10.2)
+      kramdown (~> 2.1)
+      kramdown-parser-gfm (~> 1.0.1)
+      mustache (~> 1.0)
+      nokogiri (~> 1.10, >= 1.10.10)
     octokit (6.0.1)
       faraday (>= 1, < 3)
       sawyer (~> 0.9)
@@ -217,11 +211,11 @@ DEPENDENCIES
   middleman-search!
   middleman-syntax
   nokogiri (~> 1.13)
+  nronn
   octokit (~> 6.0)
   pry
   pry-byebug
   rake
-  ronn-ng!
   rubocop
 
 RUBY VERSION
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ gem "rake"
 gem "faraday-retry", "~> 2.0"
 gem "octokit", "~> 6.0"
 ## To generate ERB files from ronn files from rubygems/rubygems
-gem "ronn-ng", github: "deivid-rodriguez/ronn-ng", branch: "fix-charset"
+gem "nronn"
 ## To strip (man:strip_pages)
 gem "nokogiri", "~> 1.13"
 


olleolleolle (approved):
Awesome find, this simplifies a lot.

deivid-rodriguez:
I think it's a bit unfair to not mention ronn-ng at all here, since it's ronn-ng which took care of removing this dependency duplication. I think you should try to get ronn-ng mantainer to hand over maintenance to you, to save the whole community yet another migration to a different gem (I think OS packagers are already using ronn-ng).

If you don't succeed on taking over maintenance, then I think you should at least credit ronn-ng in this PR and in nronn's README. Also, I'm not sure why you didn't create nronn as a fork of ronn-ng, since that's what it is.

tnir (commented):
Found a regression, pushing back to draft.

deivid-rodriguez:
We're getting segfaults with ronn-ng now for some reason. Maybe nronn fixes them? What was the regression that you found, @tnir? The new fork looked pretty great. I think as long as we give proper attribution to ronn-ng, we could move forward with it!

tnir:
@deivid-rodriguez Thanks. Were you looking at [this failure](https://github.com/rubygems/bundler-site/actions/runs/3776821410/jobs/6420303837) only in GitHub Actions environment? I was not able to find failures even with `ronn-ng` on my local.

- Failed job: https://github.com/rubygems/bundler-site/actions/runs/3776821410/attempts/1
- and re-run succeeded job: https://github.com/rubygems/bundler-site/actions/runs/3776821410

tnir:
Anyway merge conflicts are now all resolved.

deivid-rodriguez:
@tnir Yes, I was looking at those failures indeed. Sometimes it does work, but they seem more frequent now. Since the crash happens deep inside hpricot's gem code, I figure moving away from that gem would do the trick.

[Hpricot](https://github.com/hpricot/hpricot) hasn't received updates for almost 10 years and declares itself as dead in its README, recommending to move on to nokogiri like nronn does.

I had a look at the site built from this PR and everything looks fine!

deivid-rodriguez:
@tnir Would you be willing to take over [ronn-ng](https://github.com/apjanke/ronn-ng) maintenance? If current maintainer is ok with that, that'd be probably best for the community.

Gemfile.lock
```diff
@@ -159,10 +165,6 @@ GEM
       pairing_heap (>= 0.3.0)
       rexml (~> 3.2, >= 3.2.4)
       stream (~> 0.5.3)
-    ronn (0.7.3)
-      hpricot (>= 0.8.2)
-      mustache (>= 0.7.0)
-      rdiscount (>= 1.5.8)
```
deivid-rodriguez, line 165: I just noticed that the `rdiscount` dependency has been removed by this PR, yet it's still included in the `Gemfile.lock`. Do you know how that happened?

tnir: @tnir Just my laziness caused this problem 🙇 When I resolved the merge conflicts, I made a mistake not to remove the line of `rdiscount` dependency. The latest commit resolved the problem.

deivid-rodriguez: I see, it's not a problem at all.

I believe older Bundler versions used to clean this up automatically. But I removed that cleanup pass for performance. Will look into adding back that cleanup to a "non performance-critical" code path. In any case, it's not a problem, I believe Bundler can deal with such `Gemfile.lock` just fine, ignoring the completely ignoring the extra dependency.

deivid-rodriguez (approved):
@tnir Approving this since I don't think it should block this PR, but if you can confirm that you'd be willing to take over ronn-ng maintenance, I can ask upstream.

deivid-rodriguez:
Actually I'm doubly confused now that I double checked dependencies, since `ronn-ng` has also dropped the dependency on [both nokogiri and rdiscount](https://github.com/apjanke/ronn-ng/blob/5c3da4065a4d3c42d80b3751cf5350a8c42bdc32/ronn-ng.gemspec). What's the difference between `nronn` and `ronn-ng`?

PR #858, opened 2022-08-18
Use h1 in bundle plugin

### LINKED ISSUE #832: Use `h1` in `bundle plugin`

tnir:

Headings in `bundle plugin` (https://bundler.io/v2.3/bundle_plugin.html for v2.3) are `h2` tag instead of `h1`, which is the standard formatting.

Also meta description is missing.

Closes #832

Replace h2 with h1 in the command title.

Add meta description for SEO.

Use h1 in bundle plugin (only v2.1 through v2.3).

We could migrate this doc to rubygems/rubygems, but I do not think it is so easy to do so against v2.1 and v2.2.

### DIFF
--- a/source/v2.1/bundle_plugin.html.haml
+++ b/source/v2.1/bundle_plugin.html.haml
@@ -1,4 +1,8 @@
-%h2 bundle plugin
+---
+title: bundle plugin
+description: bundle-plugin allows you to manage Bundler plugins.
+---
+%h1 bundle plugin
 
 .contents
   .bullet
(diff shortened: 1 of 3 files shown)


tnir (commented):
Umm,,, it is still invalid to have `description` in the frontmatter to add `meta` name=description as per
https://github.com/rubygems/bundler-site/blob/9d1d23d699cdbb99e721e3409818a6b445bf7e5f/source/layouts/base.haml

At this moment, I am merging this and will fix it later

PR #860, opened 2022-08-18
Rewrite plugin manual

### LINKED ISSUE #859: bundle plugin uninstall should be added (v2.2+)
https://bundler.io/v2.1/bundle_plugin.html announced:

> In 2.2.0, an additional uninstall command will be added.

but https://bundler.io/v2.2/bundle_plugin.html and https://bundler.io/v2.3/bundle_plugin.html had no change.

## TODO

- [x] #869

tnir:
### Reviewer guide
Check out only `source/v2.3/bundle_plugin.html.haml` when reviewing. v2.2 and v2.3 are identical and v2.1 is very similar to them.

- `bundle plugin uninstall` sub-command is missing in bundler plugin v2.2+.
- Content is not very precise.
- Formatting is rude.

Closes #859

- Adds `uninstall` sub-command to `bundler plugin` man (v2.2 and v2.3).
- Every option is verified.
- Formatting is now aligned with man pages based on ronn-generated HTML.

### DIFF
--- a/source/v2.3/bundle_plugin.html.haml
+++ b/source/v2.3/bundle_plugin.html.haml
@@ -2,53 +2,83 @@
 title: bundle plugin
 description: bundle-plugin allows you to manage Bundler plugins.
 ---
-%h1 bundle plugin
-
 .contents
-  .bullet
-    .description
-      Manage Bundler plugins.
-    .notes
-      %ul
-        %li
-          See also
-          = link_to 'How to write a Bundler plugin', './guides/bundler_plugins.html'
-        %li
-          In 2.2.0, an additional <code>uninstall</code> command will be added.
-    :code
-      $ bundle plugin install PLUGINS
-    .description
-      Install the plugin from the source
-    .notes
-      %p
-        Options:
-      %p
-        <code>--source</code>: URL of the RubyGems source to fetch the plugin from
-      %p
-        <code>--version</code>: The version of the plugin to fetch
-      %p
-        <code>--git</code>: URL of the git repo to fetch from.
-      %p
-        <code>--local-git</code>: Path of the local git repo to fetch from. This is the
-        = link_to 'recommended', 'https://github.com/rubygems/bundler-site/pull/464'
-        way to install a plugin from a local filesystem (a useful technique for plugin developers)
-      %p
-        <code>--branch</code>: The git branch to checkout
-      %p
-        <code>--ref</code>: The git revision to check out
+  %h1 bundle plugin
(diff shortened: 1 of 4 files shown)


tnir:
https://github.com/rubygems/bundler-site/runs/7904983470?check_suite_focus=true fails.
> [W] IdNames: `id` attribute must be in lisp-case

We might have to disable `IdNames` in HAML-Lint inline to align with ronn-based HTMLs.

source/v2.3/bundle_plugin.html.haml
```diff
@@ -1,50 +1,80 @@
...
+            %code http[s]://host.xz[:port]/path/to/repo.git
+          %li
+            %code /path/to/repo
+          %li
+            %code file:///path/to/repo
       %p
-        <code>--ref</code>: The git revision to check out
+        When you specify <code>--git</code>/<code>--local-git</code>, you can use <code>--branch</code>
+        or <code>--ref</code> to specify any branch, tag, or commit hash (revision) to use.
+        When you specify both, only the latter is used.
```
tnir, line 60: **note**: This behavior looks like a bug to me. Not sure if we really need to use two different options.

`git-clone -b` supports for arbitrary `rev` value.
https://git-scm.com/docs/git-clone#Documentation/git-clone.txt--bltnamegt

> -b \<name\>
> --branch \<name\>
> Instead of pointing the newly created HEAD to the branch pointed to by the cloned repository’s HEAD, point to `<name>` branch instead. In a non-bare repository, this is the branch that will be checked out. `--branch` can also take tags and detaches the HEAD at that commit in the resulting repository.

tnir: Filed as https://github.com/rubygems/rubygems/issues/5845

deivid-rodriguez (commented):
I think command documentation was imported from upstream. Should we create a man page upstream instead?

tnir:
Do you mean other command references were already imported to the upstream (rubygems/rubygems)? If yes, I will create a new man page for bundle-plugin(1) there. But how about ones for v2.2 or earlier? Do you like to backport the one to the maintenance branches?

deivid-rodriguez:
Yes, creating a man page upstream is definitely the way to go because it will not only help the doc site but also the CLI help. 

For older versions, I'm not 100% sure, I guess we can backport to maintenance branches although without any guarantees of ever releasing, but at least they can be pulled by the docs site?

tnir:
Closing in favor of https://github.com/rubygems/rubygems/pull/5853

PR #828, opened 2022-07-31
Clean up whats new pages

tnir:
Moves the navigation to what's new pages in other versions to bottom and adds the `h1` heading titled as `What's New in vX.Y`.

Closes #827

| before | after |
|-|-|
| bundler io_v1 12_whats_new html | 4567-rubygems-bundlersite-xaimk8cdg5n ws-us58 gitpod io_v1 12_whats_new html |

Signed-off-by: Takuya Noguchi [takninnovationresearch@gmail.com](https://github.com/sponsors/tnir)

PR #873, opened 2022-08-26
Integrate links to whats_new into the sidebar of command references

### LINKED ISSUE #629: Replace guides in sidebar with command references for each Bundler version for each "what's new page"
From https://github.com/rubygems/bundler-site/pull/625#pullrequestreview-1028327402:

@deivid-rodriguez says:
> Replace the sidebar for each "what's new page" to instead show the documentation sidebar for each version (the one here: `https://bundler.io/v2.3/man/bundle-install.1.html`).

From https://github.com/rubygems/bundler-site/pull/828#issuecomment-1200887751

> This looks good, but I was thinking we could maybe merge release notes into "Reference Guides".
> 
> Something like this
> 
> Captura de Pantalla 2022-08-01 a las 10 28 48
> 
> And we can completely get rid of the release notes version switcher in favor of the sidebar version switcher.

tnir:

"What's new" in each version (e.g., `https://bundler.io/v2.3/whats_new.html`) looks orphan from the command references and it has a different navigation based on buttons (see #828) from the command references (e.g., `https://bundler.io/v2.3/man/bundle-install.1.html`).

Closes #629

We can unify these two navigations in the side bar appeared in the command references.

- Adds a link to "What's new" into the side bar in the command references.
- Changes the navigation "What's new" in between versions from the buttons to the dropdown-based one existing in the command references. 
- Changes the number of columns in `docs.html` from three to two by removing the dedicate column for What's new.
- Removes the old partial template file and references to it from `whats_new.html` in each version.
- Reduces the differences in `source/v:ver/docs.html.haml` files to improve the maintainability, but remains some links in the older versions before v1.15. e.g.:
    ```diff
    $ diff -u source/v{1.14,1.15}/docs.html.haml
    --- source/v1.14/docs.html.haml 2022-08-26 00:00:00.000000000 +0000
    +++ source/v1.15/docs.html.haml 2022-08-26 00:00:00.000000000 +0000
    @@ -16,8 +16,6 @@
        %ul.ul-padding
            - guides.each do |guide|
            %li= link_to_guide guide
    -        -# The below entry is not included in the above `guides`
    -        %li= link_to 'Compatibility', '/compatibility.html'
    
        .col-md-5.offset-md-1.col-12
        %h3
    ```

#### Screenshots

| page | before | after |
|-|-|-|
| docs.html |  | docs |
| bundle-install.1.html | install-b2 | install |
| whats_new.html | whatsnew_b | whatsnew |

Other options are not considerable.


deivid-rodriguez (commented):
Looks good!

The only thing that feels off to me is that with the new `docs.html` layout, the "Reference Guides" section header feels a bit confusing and I think we could maybe get rid of it.

This is how it looks now:

Captura de Pantalla 2022-08-26 a las 9 23 22

"Reference Guides" is clearly a column header that explains what's below.

This is how it looks with this PR:

Captura de Pantalla 2022-08-26 a las 9 25 57

Since typography of "Release Notes" above (and the link it includes "What's new in 2.3") is exactly the same, I think it could be misinterpreted that the content of the "Reference Guides" section is just a link named "In-depth articles with information on each primary command and utility, and how to use them." that links to these guides.

I think it would be more clear if we remove this part completely, and leave it like this

Captura de Pantalla 2022-08-26 a las 9 33 19

tnir:
Thanks @deivid-rodriguez. Another idea on `docs.html` might be to remove the link to "What's new in vX.Y" completely from `docs.html` since we have the link in each command reference page. What do you think it instead?

deivid-rodriguez:
That sounds good to me too. Whatever you like better would be ok with me.

deivid-rodriguez (approved):

PR #876, opened 2022-08-26
Enable HtmlAttributes for all files in HAML-Lint

tnir:

Some linters were disabled when HAML-Lint was introduced in #776.

`HtmlAttributes` HAML linter can be enabled for all files.

Removes exclusion of files with lints and enables the linter.

n/a


tnir (commented):
Merging this while some `meta`s in `head` are now obsolete.

PR #877, opened 2022-08-26
Enable RepeatedId for all files in HAML-Lint

tnir:

`RepeatedId` linter in HAML-Lint was not enabled in #776.

#863 addressed lints against `RepeatedId`.

Enables `RepeatedId` for all files in HAML-Lint.

n/a


tnir (commented):
HAML-Lint passed even in CI.

PR #718, opened 2022-07-20
Remove unnecessary year/month/day blog indexes

tnir:
### What was the end-user problem that led to this PR?

Unnecessary year/month/day blog indexes have happens after #122.

Closes #713

### What was your diagnosis of the problem?

Remove (about 90) unnecessary and meaningless index pages to improve SEO.

### What is your fix for the problem, implemented in this PR?

Reverts (almost) all of #122 and adds the ephemeral redirections (maybe 3-6 months) for dead URLs for (public) search engines.

### Why did you choose this fix out of the possible options?

I cannot imagine another option after #122. Per redirection rules, we cannot use glob (list files) in `/source/blog/20*.html.markdown` since new articles in future do not require redirections.

Signed-off-by: Takuya Noguchi [takninnovationresearch@gmail.com](https://github.com/sponsors/tnir)

PR #899, opened 2022-08-30
Middleman redirects require index.html in the old paths

tnir:
https://middlemanapp.com/basics/redirects/ says:

> To generate a redirect in Middleman, add the following to `config.rb`:
> 
> ```
> redirect "my/old/path.html", to: "/my/new/path.html"
> ```
> 
> The old path should not have a leading slash and should refer to a file such as `path/index.html`, not just a path.

Fixes up #718


tnir (commented):
🙏

Issue #553, opened 2021-01-15
Move all urls from github.com/bundler/bundler to github.com/rubygems/rubygems.


simi:
There is a lot of files using old bundler source path, which is being redirected to archived repo (rubygems/bundler).

### example
at https://bundler.io/guides/bundler_plugins.html there is link to  [bundler/bundler plugin/events.rb](https://github.com/bundler/bundler/blob/master/lib/bundler/plugin/events.rb), but it should really link to [rubygems/rubygems plugin/events.rb](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/plugin/events.rb)

I was about to fix this one case, but I have found out it is not the only page having this problem. I would rather fix this at all files at once.

```
ag -l github.com/bundler/bundler | wc -l
91
```

tnir:
I am taking this stale issue. First we have 60 occurrences now:
```
$ git grep github.com/bundler/bundler | wc -l
60
```

PR #924, opened 2022-09-13
Replace links to bundler/bundler with rubygems/rubygems/bundler

tnir:
The links updated in this change refer to master branch, namely the latest version of the document and/or code.

- Second part of #553


.haml-lint.yml
```diff
@@ -14,7 +14,7 @@ linters:
       - "source/layouts/_navbar.haml"
 
   LineLength:
-    max: 849
+    max: 859
```
deivid-rodriguez, line 17: In my opinion, line length cops get in the middle too much. I would disable it, but it's your call!

tnir: Totally agree. Addressing this in #926 before this PR! 🍀

Issue #928, opened 2022-09-14
Unify markup to Markdown (in this repo)


tnir:
# What was the ~~end-user~~ contributor problem that led to this PR?

Middleman supports for multiple template engines as shown in https://middlemanapp.com/basics/templating-language/, and so we had used three engines: HAML, Markdown in addition to ERB for files from the upstream's `ronn` man pages.

Conversion from Markdown to HAML was done in #230 in 2016 😭 .

# What was your diagnosis of the problem?

As HAML allows us to flexibly use Ruby code in their files, these files tend to be complicated. And so line wrapping gets harder to use, which results in problems recently resolved in #926.

As discussed in #829, `ronn` files in `rubygems/rubygems/bundler/..../man` are based on Markdown.

While HAML provides helpers to easily avoid duplications, there have been many duplications (see https://github.com/rubygems/bundler-site/milestone/2 for detail).

After completion of https://github.com/rubygems/bundler-site/milestone/2, we will have three types of files in this repository: guides, navigations, blog articles.

## guides
While HAML files are dominant, it looks like the existing HAML can be converted to markdown.
```
$ git ls-files source/guides/*.haml source/localizable/guides/*.haml
source/guides/bundler_plugins.html.haml
source/guides/bundler_setup.html.haml
source/guides/bundler_sharing.html.haml
source/guides/bundler_workflow.html.haml
source/guides/deploying.html.haml
source/guides/faq.html.haml
source/guides/gemfile.html.haml
source/guides/gemfile_ruby.html.haml
source/guides/getting_started.html.haml
source/guides/git.html.haml
source/guides/groups.html.haml
source/guides/rails.html.haml
source/guides/rationale.html.haml
source/guides/rubygems.html.haml
source/guides/rubymotion.html.haml
source/guides/sinatra.html.haml
source/guides/updating_gems.html.haml
source/localizable/guides/bundler_setup.es.html.haml
source/localizable/guides/bundler_sharing.es.html.haml
$ git ls-files source/guides/*.md source/localizable/guides/*.md
source/guides/bundler_2_upgrade.html.md
source/guides/bundler_docker_guide.html.md
source/guides/bundler_in_a_single_file_ruby_script.html.md
source/guides/git_bisect.html.md
source/guides/rubygems_tls_ssl_troubleshooting_guide.html.md
source/localizable/guides/creating_gem.en.html.md
source/localizable/guides/creating_gem.pl.html.md
[...]

PR #929, opened 2022-09-14
PoC: Migrate HAML files to Markdown files

tnir:

See #928

All HAML files can be (manually) converted to Markdown.

One HAML file was manually converted to Markdown.

`kramdown` Markdown parser does not support for triple-backticks for multiple-line code blocks, so we need to use triple-tildes instead.


tnir:
Closing in favor of #1064 (also see #928)

Issue #618, opened 2022-06-23
/man/.* should be redirected to /v2.3/man/.*


tnir:
https://bundler.io/man/bundle-install.1.html is now the man for `bundle install` on Bundler 2.3 as of writing, but with release of 2.4, the content of `/man/bundle-install.1.html` will be updated.

I think `/man/bundle-install.1.html` should be redirect to `/v2.3/man/bundle-install.1.html` (if the current version is v2.3).

gustavothecoder:
Can I work on this issue?

deivid-rodriguez:
Yes, please!

deivid-rodriguez:
Closed by #965.

tnir:
Reopened by #990

tnir:
Closing in favor of #992 while we tried it in #965, which was reverted in #990. Note that #992 would be resolved sooner in #993.

PR #965, opened 2022-12-09
Redirect to the latest man page when version isn't specified

gustavothecoder:

 #618

My fix redirects all requests for `/man/*.html` to `/<latest_bundler_version>/man/*.html` using `redirect`.

I chose this fix because it follows the project standards and is very simple.

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -165,13 +165,21 @@ end
   redirect "v#{version}/guides/bundler_2_upgrade.html", to: "guides/bundler_2_upgrade.html"
 end
 
+# Redirect to the latest man page when version isn't specified
+man_files = Dir.glob("./source/#{config[:current_version]}/man/*.html.erb")
+man_files.each do |file|
+  url = file.delete_prefix("./source/#{config[:current_version]}/").delete_suffix(".erb")
+  latest_man = "#{config[:current_version]}/#{url}"
+  redirect url, to: latest_man
+end
+
 redirect "sponsors.html", to: "https://rubygems.org/pages/sponsors" # Backwards compatibility
 
 page "/conduct.html", layout: :two_column_layout
 page "/compatibility.html", layout: :two_column_layout
 page /\/v(\d+.\d+)\/(?!bundle_|commands|docs|man)(.*)/, layout: :two_column_layout
 page /\/v(.*)\/man\/(.*)/, layout: :two_column_layout
-page /\/man\/(.*)/, layout: :two_column_layout
+page /\/man\/(.*)/, layout: false
 page /\/v(.*)\/guides\/(.*)/, layout: :two_column_layout
 page /guides\/(.*)/, layout: :two_column_layout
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Thanks @gustavothecoder! CI is failing, can you have a look?

config.rb
```diff
@@ -165,13 +165,24 @@
...
+end
+
 redirect "sponsors.html", to: "https://rubygems.org/pages/sponsors" # Backwards compatibility
 
 page "/conduct.html", layout: :two_column_layout
 page "/compatibility.html", layout: :two_column_layout
 page /\/v(\d+.\d+)\/(?!bundle_|commands|docs|man)(.*)/, layout: :two_column_layout
 page /\/v(.*)\/man\/(.*)/, layout: :two_column_layout
-page /\/man\/(.*)/, layout: :two_column_layout
+page /\/man\/(.*)/, layout: false
```
gustavothecoder, line 185: @deivid-rodriguez Now the build is passing 🎉 .

To fix the build, I re-added the man page generic path and removed the layout:
```ruby
page /\/man\/(.*)/, layout: false
```

config.rb
```diff
@@ -165,13 +165,24 @@
...
 end
 
+# Redirect to the latest man page when version isn't specified
+man_files = Dir.glob("./source/#{config[:current_version]}/man/*.html.erb")
+generic_man_urls = man_files.map do |file|
+  file.delete_prefix!("./source/#{config[:current_version]}/")
+  file.delete_suffix!(".erb")
+end
+generic_man_urls.each do |url|
+  latest_man = "#{config[:current_version]}/#{url}"
```
deivid-rodriguez, line 175: Does this simplification make sense?

```suggestion
  url = file.delete_prefix("./source/#{config[:current_version]}/").delete_suffix(".erb")
  latest_man = "#{config[:current_version]}/#{url}"
```

gustavothecoder: Yeah, a lot better.

deivid-rodriguez (approved):

deivid-rodriguez:
Merging this. Thanks for this great fix @gustavothecoder, it helps contributors to Bundler documentation a lot! 💪

tnir:
Thanks @gustavothecoder for improving the site, but it does not looks like what I meant..., for example https://bundler.io/man/bundle-init.1.html is now in plain text...

## https://bundler.io/man/bundle-init.1.html

bundler io_man_bundle-init 1 html

deivid-rodriguez:
Nice catch @tnir. @gustavothecoder could you have a look at that?

tnir:
@deivid-rodriguez @gustavothecoder Can I revert this as this introduced the bug shown as above?

gustavothecoder:
> Nice catch @tnir. @gustavothecoder could you have a look at that?

Yes, I can.

> @deivid-rodriguez @gustavothecoder Can I revert this as this introduced the bug shown as above?

I think it's a good idea.

PR #975, opened 2022-12-22
Bump ruby and bundler

deivid-rodriguez:

The problem was none. Just taking the chance to upgrade these.


olleolleolle (approved):
✨

PR #979, opened 2022-12-22
Fix some haml-lint offenses

deivid-rodriguez:

The problem was that when adding a new minor version to the documentation site, we need to remember to add exclusions to haml-lint configuration or else haml-lint will fail.

My diagnosis was that we can fix the haml-lint offenses, to get cleaner templates and also not have to remember to add new exclusions when we add new versions.

My fix is to fix the offenses and the template.

I chose this fix because just adding new exclusions was inferior in terms of maintainability.

### DIFF
--- a/.haml-lint.yml
+++ b/.haml-lint.yml
@@ -21,18 +21,6 @@ linters:
     ignored_cops:
       - "Style/StringLiterals"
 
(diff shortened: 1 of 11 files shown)


olleolleolle (approved):
Looking clear and excellent.

PR #980, opened 2022-12-22
Fix octokit getting rate-limited in CI

deivid-rodriguez:

The problem was that sometimes CI can fail due to API calls to the GitHub API getting rate-limited.

My diagnosis was, thanks to the helpful error message, that we should be doing authenticated calls.

My fix is to use the `GITHUB_TOKEN` available in CI to authenticate the API calls.


olleolleolle (approved):
Excellent, proven to work, in the other PR!

PR #984, opened 2022-12-25
Update Ruby from 3.1.3 to 3.2.0

tnir:

Now we are not on the latest Ruby as Ruby 3.2.0 was just released today: https://www.ruby-lang.org/en/news/2022/12/25/ruby-3-2-0-released/.

We are still using Ruby 3.1.

Use Ruby 3.2.0 instead of 3.1.x.

n/a

- Previous update: #975
- Blocked by #999


tnir (commented):
middleman-search does not support Ruby 3.2.0 cc @deivid-rodriguez

deivid-rodriguez:
Yep, I noticed!

I think in the case of middleman-search, it's fair to create our own middleman-search fork and start maintaining it for now, since the upstream gem seems dead.

We need to add Ruby 3.2.0 compatibility to it either way.

tnir:
@deivid-rodriguez Merging this as no regression was found.

PR #986, opened 2022-12-25
Fix rate limit when building site for deployment

deivid-rodriguez:

The problem was the deployment task can fail due to GitHub API rate limiting due to the same issue as #980.

My diagnosis was the we should also use authenticated API calls when deploying.

My fix is to apply the same solution as #980.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -39,7 +39,10 @@ jobs:
         id: pages
         uses: actions/configure-pages@v2
 
-      - run: bundle exec rake build # Output to ./_site as build_dir is configured in config.rb
+      - name: Build site
+        run: bundle exec rake build # Output to ./_site as build_dir is configured in config.rb
+        env:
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
👍 Good fix, good find.

tnir:
@deivid-rodriguez Thanks for fixing it, but I am not sure if we really need to do this. While I just guest you see some failures at https://github.com/rubygems/bundler-site/actions/workflows/deploy.yml, we can retry to fix it on failed jobs by `Re-run jobs`...

- First attempt (failed): https://github.com/rubygems/bundler-site/actions/runs/3776333752/attempts/1
- and re-run attempt (succeeded): https://github.com/rubygems/bundler-site/actions/runs/3776333752

deivid-rodriguez:
o we agree that if it fails sporadically, then there's an issue? I would rather not have to retry jobs at all to get the site deployed.

Maybe there's some issue with this fix, but it seems easy enough to wait and see if the rate limit errors reappear?

PR #989, opened 2022-12-29
Migrate from `ronn` to `ronn-ng`

deivid-rodriguez:

The problem was that the Ronn gem and its dependencies have been abandoned for 10 years. It has worked pretty well so far but we're seeing some unstability now. In particular, we've seen some C-level crashes inside the hpricot dependency.

My diagnosis was that we should migrate to newer alternatives.

My fix is to migrate to the currently maintained fork, [ronn-ng](https://github.com/apjanke/ronn-ng). I also needed to pick up [one fix](https://github.com/n-ronn/nronn/commit/09be8af91dcb01693848a9884d66a011235ce6ac) by @tnir.

I will propose this fix to ronn-ng and ask whether they would allow us to get commit and release access there.

I chose this fix because creating [yet another fork](https://github.com/n-ronn/nronn) as in #841 seems worse for the community so we could try the existing fork first.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ gem "rake"
 gem "faraday-retry", "~> 2.0"
 gem "octokit", "~> 6.0"
 ## To generate ERB files from ronn files from rubygems/rubygems
-gem "ronn"
+gem "ronn-ng", github: "apjanke/ronn-ng"
 ## To strip (man:strip_pages)
 gem "nokogiri", "~> 1.13"
 
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ gem "rake"
 gem "faraday-retry", "~> 2.0"
 gem "octokit", "~> 6.0"
 ## To generate ERB files from ronn files from rubygems/rubygems
-gem "ronn-ng", github: "apjanke/ronn-ng"
+gem "ronn-ng", github: "deivid-rodriguez/ronn-ng", branch: "fix-charset"
 ## To strip (man:strip_pages)
 gem "nokogiri", "~> 1.13"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,14 +1,4 @@
 GIT
-  remote: https://github.com/apjanke/ronn-ng.git
-  revision: 5c3da4065a4d3c42d80b3751cf5350a8c42bdc32
-  specs:
-    ronn-ng (0.10.1.pre1)
-      kramdown (~> 2.1)
-      kramdown-parser-gfm (~> 1.0.1)
-      mustache (~> 1.0)
-      nokogiri (~> 1.11, >= 1.11.0)
-
-GIT
   remote: https://github.com/deivid-rodriguez/middleman-search.git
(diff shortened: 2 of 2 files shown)


tnir (commented):
`ronn-ng` does not support Ruby 3.0+ as far as I remember, so I guess we need to spend time to maintain it. 😢

deivid-rodriguez:
I tried locally with Ruby 3.1.3 and everything worked fine. Also I couldn't pin point any commit that fixed anything regarding Ruby 3 at https://github.com/n-ronn/nronn/commits/main? Can you double check? We'll need to spend more time to maintain it if we use a fork maintained only by ourselves, rather than a community fork, don't you think?

deivid-rodriguez:
I also asked for commit and release access to ronn-ng maintainer at https://github.com/apjanke/ronn-ng/issues/82. 

If we confirm this is working fine, we can go with this and wait for a response a https://github.com/apjanke/ronn-ng/issues/82. If we get no response, or a negative one, we can switch to nronn.

How does that sound?

deivid-rodriguez:
Mmmm, the review app seems indeed broken... 🤔

deivid-rodriguez:
Actually, no, it seems fine now!

tnir:
I see, let us wait for the response for a while 👍

tnir:
But I personally do not want to use Git-based `gem` DSL since it causes to download much more than static gem...

deivid-rodriguez:
Does that cause any issues? Also, with Bundler 2.4.1 it should not download that much more since it now uses shallow clones. We're also using that for other gems, and never caused any issues that I know of, it's a good temporary stop gap.

Anyways, if you feel very strongly about that for some reason, I guess we can use your nronn gem for now. I prefer that to keep getting C-level segfaults sporadically 😅.

tnir:
Also like https://github.com/rubygems/bundler-site/pull/984#pullrequestreview-1232423769, only the personal maintainer who created the fork can update code with breaking compatibility problem.

deivid-rodriguez:
> Also like https://github.com/rubygems/bundler-site/pull/984#pullrequestreview-1232423769, only the personal maintainer who created the fork can update code with breaking compatibility problem.

I can create my middlesearch fork under the rubygems org, if that's what you mean? Or you can open a PR to my fork. I don't understand very well how that relates to using git gems or not 🤔.

deivid-rodriguez:
If there are no new objections I will merge this later today, based on what I explained above. We can switch to nronn later if ronn-ng gets confirmed as abandoned.

simi (approved):

deivid-rodriguez:
Thanks @simi!

deivid-rodriguez:
It seems ronn-ng maintainer does not want to add anybody to collaborate there. But the PR that we needed has been merged and there should be a release soon, so no big deal.

That said, I did check download numbers of ronn-ng, and they are way lower than I thought. nronn has 2K downloads, ronn-ng has 50K, and ronn has 2M. So I think forking-off would not be as disruptive to the community as I was expecting. But let's stick to ronn-ng for now and see how it goes!

PR #990, opened 2022-12-29
Revert "Redirect to the latest man page when version isn't specified"

tnir:
Reverts #965 as it simply caused the bug.

Reopens #990

- See https://github.com/rubygems/bundler-site/pull/965#issuecomment-1357282797

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -165,21 +165,13 @@ end
   redirect "v#{version}/guides/bundler_2_upgrade.html", to: "guides/bundler_2_upgrade.html"
 end
 
-# Redirect to the latest man page when version isn't specified
-man_files = Dir.glob("./source/#{config[:current_version]}/man/*.html.erb")
-man_files.each do |file|
-  url = file.delete_prefix("./source/#{config[:current_version]}/").delete_suffix(".erb")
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
I think the bug introduced by #965 is less important than the one introduced.

The bug introduced is that some pages are rendered as plain text, but I don't see these pages linked from other parts of the site, so it doesn't seem like a huge deal. Super happy to fix that of course.

The previous bug was the the "Edit this page" button would not work on pages where people actually wanted to contribute to the site. I also got this bug reported privately by folks who wanted to contribute to our documentation and got blocked by it.

So in my opinion, we should not revert #965.

tnir:
Check out https://www.google.com/search?q=bundle+install to see how big the bug #965 introduced is. If you see different, you should use https://www.google.com/search?q=bundle+install&hl=en&gl=us.

Ecosystem is more important while I am not sure how rarely you use Google search engine.

deivid-rodriguez:
I see, thanks for pointing that out!

Can we instead propose a fix for the regression? Or is it very hard to fix without breaking the "Edit this document" button again?

tnir:
Merging this first and then fix some dead links in `/man/*` if they exist.

Issue #930, opened 2022-09-17
Link for "Edit this document on GitHub" is broken


beatak:
The link on the footer "`Edit this document on GitHub`" is linking to `https://github.com/rubygems/bundler-site/blob/master/source/man/*.html.erb`. It seems the source of truth has moved to `https://github.com/rubygems/rubygems/tree/master/bundler/lib/bundler/man/*`.

deivid-rodriguez:
Thanks for reporting this, I assume you were already planning on fixing https://github.com/rubygems/rubygems/issues/5932 :)

@tnir Any ideas? I think we actually fixed this not a long time ago, I guess it broken again with the recent structure normalizations?

tnir:
@beatak Thanks for reporting.

In #615, I addressed this only in https://bundler.io/v2.3/man/bundle-exec.1.html, not in https://bundler.io/man/bundle-exec.1.html as described implicitly in https://github.com/rubygems/bundler-site/issues/620#issuecomment-1174110210.

@deivid-rodriguez I would like to address this root cause in a way described in #618.

deivid-rodriguez:
Fixed by #965!

tnir:
These links no longer exist since the top-level pages do not exist in #993.

PR #214, opened 2016-05-23
Auto-deploy after commit to master

kruczjak:
Closes #177 

Additionally I've moved some rake tasks to separate files because `Rakefile` would be quite long now.

PR #993, opened 2022-12-29
Stop making plain text pages available at the top level

### LINKED ISSUE #992: /man/.* should be redirected to /v2.4/man/.*
https://bundler.io/man/bundle-install.1.html is now the man for `bundle install` on Bundler 2.4 as of writing, but with release of 2.5 (or 3.0), the content of `/man/bundle-install.1.html` will be updated.

I think `/man/bundle-install.1.html` should be redirect to `/v2.4/man/bundle-install.1.html` (if the current version is v2.4).

deivid-rodriguez:

The problem was some pages render as plain text.

Link for "Edit this document on GitHub" is broken in top-level `/man/` pages (#930).

My diagnosis was that we should get rid of 

```ruby
page /\/man\/(.*)/, layout: false
```

from our config, since that's what's rendering content without a layout.

My fix is to remove that line and fix the issues that come up after doing that.

I chose this fix because it fixes #992 without regressing #930.

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -48,8 +48,8 @@ activate :external_pipeline,
 
 # Make documentation for the latest version available at the top level, too.
 # Any pages with names that conflict with files already at the top level will be skipped.
-Dir.glob("./source/#{config[:current_version]}/**/*").select{ |f| File.file?(f) }.each do |file_path|
-  file_path = file_path.sub(/(\.haml$|\.md$)/, "")
+Dir.glob("./source/#{config[:current_version]}/*.haml").each do |file_path|
+  file_path = file_path.sub(/\.haml$/, "")
 
   page_path = file_path.sub(/^\.\/source\//, "")
   proxy_path = file_path["./source/#{config[:current_version]}/".length..-1]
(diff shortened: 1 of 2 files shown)


config.rb
```diff
@@ -48,8 +48,8 @@
 
 # Make documentation for the latest version available at the top level, too.
 # Any pages with names that conflict with files already at the top level will be skipped.
-Dir.glob("./source/#{config[:current_version]}/**/*").select{ |f| File.file?(f) }.each do |file_path|
-  file_path = file_path.sub(/(\.haml$|\.md$)/, "")
+Dir.glob("./source/#{config[:current_version]}/*.haml").each do |file_path|
+  file_path = file_path.sub(/\.haml$/, "")
```
tnir, line 52: This change was missing in #651 when we (@tnir) consolidated `/:version/guides/*` 🙇

lib/tasks/man_generator/man.rake
```diff
@@ -29,8 +29,4 @@ task man: :update_vendor do
       Rake::Task["man:strip_pages"].execute("#{segments_up}/source/#{version}/man")
     end
   end
-
-  # Make man pages for the latest version available at the top level, too.
-  rm_rf "source/man"
-  cp_r "source/#{VERSIONS.last}/man", "source"
```
tnir, line 35: Happy to remove residues from #214 🎉

tnir (approved):

deivid-rodriguez:
Thanks for the quick merge!

PR #994, opened 2022-12-30
Redirect to the latest man page when version isn't specified

### LINKED ISSUE #992: /man/.* should be redirected to /v2.4/man/.*

tnir:

`https://bundler.io/man/bundle-init.1.html` is now dead again.

That is due to merging both #990 and #993.

Some parts originally prepared in #965 by @gustavothecoder should be re-introduced.

Closes #992

No other options are considerable.

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -165,6 +165,14 @@ end
   redirect "v#{version}/guides/bundler_2_upgrade.html", to: "guides/bundler_2_upgrade.html"
 end
 
+# Redirect to the latest man page when version isn't specified
+man_files = Dir.glob("./source/#{config[:current_version]}/man/*.html.erb")
+man_files.each do |file|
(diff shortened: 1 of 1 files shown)


tnir (commented):
`https://bundler-site-follow-up--trdjh0.herokuapp.com/man/bundle.1.html` is now correctly redirected to `https://bundler-site-follow-up--trdjh0.herokuapp.com/v2.4/man/bundle.1.html`

PR #999, opened 2022-12-31
Update middleman-search to edge to use Ruby 3.2

tnir:

Current version of middleman-search does not support Ruby 3.2 as its syntax is too legacy.

Ruby 3.2-ready code is available at https://github.com/manastech/middleman-search/pull/39 or https://github.com/tnir/middleman-search/tree/edge.

Makes the code ready to use Ruby 3.2.

Once https://github.com/manastech/middleman-search/pull/39 is merged, we can use the upstream.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ ruby File.read(File.expand_path("../.ruby-version", __FILE__)).strip
 gem "middleman", github: "middleman/middleman", ref: "038f4f606ff6cfc75cbe5c8690c5eae2e34f78f3"
 ## Extensions
 gem "middleman-blog"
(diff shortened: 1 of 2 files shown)


tnir (commented):
Search feature does not degraded in review apps.

tnir:
@deivid-rodriguez Let me know how to work well in your fork as per [your concept](https://github.com/rubygems/bundler-site/pull/989#issuecomment-1367264817) while I am inclined to use Ruby 3.2 quickly with merging.

deivid-rodriguez:
Nice, works for me @tnir and I believe "edge" is a good name. Although I'm not sure it makes sense to point at any pending PRs, since we have several PRs pending now stacked on top of each other. In my opinion, next step would be to choose a name for a fork and publish it to rubygems.org.

PR #1006, opened 2023-01-01
Rewrite guides/bundler_workflow from HAML to Markdown

tnir:

Part of #928 for `source/guides/bundler_workflow.html.*`

Source of `/guides/bundler_workflow.html` can be rewritten in Markdown with HTML snippets.

Replaces guides/bundler_workflow from HAML to Markdown.

| before | after |
|-|-|
| bundler io_guides_bundler_workflow html | bundler-site-tnir-928-g-s3vf5a herokuapp com_guides_bundler_workflow html |

Less indents are king while three backticks cannot be used (use three tildes instead 〰️ )


tnir (commented):
Very small visual changes can be observed, but we can ignore them compared with advantages of Markdown 👑.

deivid-rodriguez:
It's cool to be able to write markdown directly 👏 🤙 

How do you access this guide, by the way? Are we linking to it internally or it's only accessible through organic search? Should we add it to the guides index?

PR #1010, opened 2023-01-01
Update kramdown-parser-gfm from 1.0.1 to 1.1.0

tnir:

We still use kramdown-parser-gfm 1.0.1, which was released in 2019 (4 years ago).

🎉 https://github.com/kramdown/parser-gfm/blob/master/CHANGELOG.md#110--2019-05-29

Simply we can use the latest published version of [kramdown-parser-gfm](https://rubygems.org/gems/kramdown-parser-gfm).

Forces to use kramdown-parser-gfm 1.1.0 instead of 1.0.1

But 1.1.0 was released in 3.5+ years ago.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -96,7 +96,7 @@ GEM
     json (2.6.3)
     kramdown (2.4.0)
       rexml
-    kramdown-parser-gfm (1.0.1)
+    kramdown-parser-gfm (1.1.0)
       kramdown (~> 2.0)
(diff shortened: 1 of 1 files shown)


tnir (commented):
`https://bundler-site-tn-kramdow-mymchc.herokuapp.com/v2.4/man/bundle-install.1.html` was perfectly rendered as before 🤩

tnir:
Build time does not change at all.

PR #1019, opened 2023-01-01
Update browserslist from 4.18.1 to 4.21.4

tnir:
browserslist is depended only by `webpack`.


tnir (commented):
This does not affect anything since [target](https://webpack.js.org/configuration/target/) or browserslist configuration has not been specified since #562.

I (or some person) will do it later, maybe in 2Q 2023.

PR #1020, opened 2023-01-01
Update terser from 5.14.2 to 5.16.1

tnir:
https://github.com/terser/terser/blob/v5.16.1/CHANGELOG.md

This allows us to use ES 2022 🎉 


tnir (commented):
Not sure when we can use ES 2022 though.

PR #1025, opened 2023-01-01
Rewrite guides/sinatra from HAML to Markdown

tnir:

Part of #928 for `source/guides/sinatra.html.*`

Source of `/guides/sinatra.html` can be rewritten in Markdown with HTML snippets.

Replaces guides/sinatra from HAML to Markdown.

~~We can squash and merge it in this PR since the content is too short 🥇~~


tnir (commented):
I will remove the line `require.*rubygems` in another PR.

PR #717, opened 2022-07-20
Add /about.html

tnir:
### What was the end-user problem that led to this PR?

Closes #716

### What was your diagnosis of the problem?

n/a

### What is your fix for the problem, implemented in this PR?

- Add /about.html with i18n.
- Add a link to the about in footer.

### Why did you choose this fix out of the possible options?

Signed-off-by: Takuya Noguchi [takninnovationresearch@gmail.com](https://github.com/sponsors/tnir)

PR #1047, opened 2023-01-02
Rewrite about from HAML to Markdown

tnir:
I was so stupid when I created #717 😭 😭 

- Part of #928


tnir (commented):
Less number of DOMs is also 👑

PR #1057, opened 2023-01-03
Rewrite deploying guide from HAML to Markdown

tnir:
This PR simply convert the page from HAML to Markdown.

- Part of #928


tnir (commented):
Visual regression is not found while the contents are too stale...

PR #1064, opened 2023-01-04
Rewrite getting_started guide from HAML to Markdown

tnir:
- Part of #928
- Replaces #929 😇 


source/guides/getting_started.html.md
```diff
@@ -0,0 +1,139 @@
...
+## What is Bundler?
+
+Bundler provides a consistent environment for Ruby projects by tracking
+and installing the exact gems and versions that are needed.
+
+Bundler is an exit from dependency hell, and ensures that the gems
+you need are present in development, staging, and production.
+Starting work on a project is as simple as `bundle install`.
+
+<a href="../whats_new.html" class="btn btn-primary">What's new in Bundler</a>
```
tnir, line 13:
```suggestion
<a href="../v2.4/whats_new.html" class="btn btn-primary">What's new in Bundler</a>
```
We need to hard-code version since we do not have the redirection from `/whats_new.html` to `/v2.4/whats_new.html` at this moment while I do not think we need to do so even in the future.

source/guides/getting_started.html.md
```diff
@@ -0,0 +1,139 @@
...
+and [RubyGems](https://rubygems.org/pages/download) installed. If you do not have Ruby
+and RubyGems installed, do that first and then check back here!
+
+Getting started with bundler is easy!
+Specify your dependencies in a Gemfile in your project's root:
+
+~~~ruby
+source 'https://rubygems.org'
+gem 'nokogiri'
+gem 'rack](~> 2.2.4'
```
tnir, line 29: See #1086 for the fix.

PR #1066, opened 2023-01-04
Comment around initial section

olleolleolle:
If someone does not remove this section, the ticket is more difficult to scan.

The problem was that editing Pull Requests can be fiddly and you want to avoid confusing the reader. The pr author opts to fill in only what they know, ignoring the rest. 

The template text in the beginning is not one of those "fill in the blanks" that the others are. 

My diagnosis was that some users leave that text section unedited, making their text read worse. 

My fix is to wrap only that section in an HTML comment, hiding it from view.

I chose this fix for the initial paragraph only because the other sections read more like sentences to complete.


deivid-rodriguez (approved):

tnir (approved):
I think it is so fiddly!

PR #1070, opened 2023-01-07
Good bye, Pry: replace it with irb

tnir:


tnir (commented):
No problem was found.

PR #1091, opened 2023-02-02
fix link to blog article

btrd:

The problem was the broken link to the blog article

My diagnosis was that the link was wrong

My fix is to use the right url

I chose this fix because it's the right one


olleolleolle (approved):
Thanks, @btrd !

PR #1102, opened 2023-03-08
Tweak compatibility page

deivid-rodriguez:

The problem was that I believe if the information is given split in two different sections, the policy becomes harder to understand.

My diagnosis was that it would be more clear if both sections are merged.

My fix is to merge both sections.

### DIFF
--- a/source/compatibility.html.haml
+++ b/source/compatibility.html.haml
@@ -3,33 +3,11 @@ title: Bundler compatibility with Ruby
 description: Ruby and RubyGems requirements needed for Bundler compatibility
 ---
 
-%h2 Bundler compatibility with Ruby
+%h2 Bundler compatibility with Ruby & RubyGems
 
 %p.contents
   The latest Bundler release should always support, at the very least, all
   rubies that have not yet reached their End of Life date.
-
-.contents
-  .bullet
-    .description
-      Bundler <b>2.4</b> or higher
-    .notes
-      %ul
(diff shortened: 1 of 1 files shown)


tnir (approved):
👍

olleolleolle (approved):
So much clearer!

source/compatibility.html.haml
```diff
@@ -3,56 +3,37 @@ title: Bundler compatibility with Ruby
...
       %ul
-        %li Requires a minimum RubyGems version of <b>3.0.1</b>.
+        %li requires a minimum ruby version of <b>2.6.0</b>, and a minimum RubyGems version of <b>3.0.1</b> (version of RubyGems shipped with ruby <b>2.6.0</b>).
 
     .description
       Bundler <b>2.0</b> through <b>2.3</b>
     .notes
       %ul
-        %li Requires a minimum RubyGems version of <b>2.5</b>, since that's the version of RubyGems included with Ruby <b>2.3.0</b>.
+        %li requires a minimum ruby version of <b>2.3.0</b>, and a minimun RubyGems version of <b>2.5</b> (version of RubyGems shipped with ruby <b>2.3.0</b>).
```
deivid-rodriguez, line 33:
```suggestion
        %li requires a minimum ruby version of <b>2.3.0</b>, and a minimum RubyGems version of <b>2.5</b> (version of RubyGems shipped with ruby <b>2.3.0</b>).
```

PR #1135, opened 2023-04-17
Update Ruby from 3.2.0 to 3.2.2

olleolleolle:

The problem was very small: a `parser` gem output about "you aren't running the same version as the `parser` gem is understanding".

My diagnosis was we are running a slightly-older Ruby.

My fix was to imitate #984 to make the same kind of upgrade.

So, now `parser` (the gem) emits no warnings on using it in `bundle exec`.

I chose this fix because it seemed small. And also, same as the previous one.


Gemfile.lock
```diff
@@ -236,7 +236,7 @@ DEPENDENCIES
   rubocop-rake
 
 RUBY VERSION
-   ruby 3.2.0p0
+   ruby 3.2.2p53
 
 BUNDLED WITH
    2.4.2
```
olleolleolle, line 242: I didn't update to a later version of Bundler. We can do that separately.

tnir:
Thanks @olleolleolle!

PR #1142, opened 2023-05-03
chore(deps): update bundler from 2.4.2 to 2.4.12

tnir:

Nothing

### What was the developer problem that led to this PR?

The developer of this site would use the latest version of Bundler as of writing.

https://rubygems.org/gems/bundler/versions/2.4.12 is the latest version of Bundler.

Update Bundler to 2.4.2 to 2.4.12 (latest).

Dogfooding might be better than nothing.


tnir (commented):
Looks perfect.

PR #1143, opened 2023-05-03
chore(deps): update postcss from 8.4.20 to 8.4.23

tnir:

postcss npm package is now stale.

postcss 8.4.23 is already available, as of writing, which includes some bugfix shown in https://github.com/postcss/postcss/releases/tag/8.4.22.

Update postcss from 8.4.20 to 8.4.23 and nanoid from 3.3.4 to 3.3.6.

Nothing considered.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1162,10 +1162,16 @@
       }
     },
     "node_modules/nanoid": {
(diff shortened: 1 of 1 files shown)


tnir (commented):
No regressions found.

PR #1150, opened 2023-05-19
Clarify that Bundler comes preinstalled by default

deivid-rodriguez:

The problem was that it was not clear that you don't really need to install Bundler if you have Ruby installed.

My fix is to make it explicit that Ruby comes with Bundler by default.


indirect (approved):
thanks!

PR #1156, opened 2023-05-29
chore(deps): update Node.js to 16.x to 18.x

tnir:

The problem was for us to stay with Node.js 16.x, which will be deprecated in less than 4 months.

My diagnosis was that there has been no version compatible with Node.js 18 in mini_racer until https://github.com/rubyjs/mini_racer/pull/261 was merged.

My fix is to update mini_racer from 0.6.3 to 0.8.0 and libv8-node from 16.10.0.0 to 18.16.0.0 to update Node.js to 16.x to 18.x.

I chose this fix because Node.js 16.x will be deprecated very soon and we need to use Node.js 20.x in the near future.


olleolleolle (approved):
Good choice, and yay, new release!

tnir (commented):
No regression confirmed in search (middleman-search) even at https://bundler-site-tn-update--akew0d.herokuapp.com/

deivid-rodriguez:
Nice!

PR #1180, opened 2023-06-26
Haml lint fix

olleolleolle:

The PR that updates haml-lint fails its lint step. #1176 

Perhaps haml-lint's rules evolved to take more details into account.

More RuboCop rules.

This PR makes haml-lint pass without failures.

- follow some of the rules
- disable one rule, which kind of does not make sense in HAML

It was quick.

### DIFF
--- a/.haml-lint.yml
+++ b/.haml-lint.yml
@@ -13,6 +13,7 @@ linters:
     enabled: true
     ignored_cops:
       - "Style/StringLiterals"
+      - "Style/IfUnlessModifier"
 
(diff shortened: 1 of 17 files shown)


indirect (approved):
looks good to me for the time being, happy to make future adjustments if @tnir believes they would be helpful.

olleolleolle:
Perfect, then we can move forward, and have the bots update the tools.

tnir:
Thanks 👍

PR #1181, opened 2023-06-26
Refactor to use #match?

olleolleolle:

None, this is a refactoring.

My diagnosis was we are not using the captures for anything, so we could use `#match?` here, which only returns a boolean.

My fix is to change the implementation.

I chose this fix because it's efficient.


hsbt (approved):

tnir (approved):
🎉

PR #1182, opened 2023-06-27
chore(deps): update Node.js to 16.x to 18.x on CI

tnir:
- Follows up PR #1156

n/a

### What was the developer problem that led to this PR?

They will not want to use Node.js 16, which will be EOL'd very soon.

CI can use Node.js 18, which is the current active LTS.

Node.js 18 is used instead of 16 even on CI.

Node.js 16 will be EOL'd in 2-3 months.


olleolleolle (approved):
To me, this looks very uncontroversial. Good fix!

PR #1186, opened 2023-07-07
Fix grammar (subject-verb agreement) in git.html.md

dubaaron:

I just noticed a small grammar error in the documentation.

Subject-verb agreement was incorrect: "... every time the local git repository change, ..."

Change the phrase to read "... every time the local git repository changes, the changes will be automatically picked up by Bundler."

I chose this fix because...
It's a straightforward fix that corrects the grammar and makes it sound correct to native English readers without changing the meaning.


olleolleolle (approved):
I think this sounds clearer.

PR #1191, opened 2023-07-11
Fix Prettier lint issue

olleolleolle:

The problem was a failed build in PR #1187. Prettier check failed.

My diagnosis was that the code should be changed to conform.

My fix was to run `npx prettier -w assets/javascripts` locally, then check again.

I chose this fix because I wanted to separate this change from any package upgrades.


olleolleolle:
OK, the other branch got to green, closing!

PR #1222, opened 2023-09-13
Update known_plugins.yml and its Rake task

### LINKED ISSUE #505: Known bundler plugins

ccutrer:
closes #505

and update the rake task. use RubyGems itself to fetch data from rubygems.org, and actually filter gems by ones that have a proper Bundler plugins.rb.

### DIFF
--- a/data/known_plugins.yml
+++ b/data/known_plugins.yml
@@ -1,21 +1,81 @@
 ---
+- :name: bootboot
+  :summary: Dualbooting your ruby app made easy.
+  :uri: https://github.com/shopify/bootboot
+- :name: bundler-as_of
+  :summary: Resolve gem dependencies as-of a date in the past.
+  :uri: https://github.com/flavorjones/bundler-as_of
 - :name: bundler-changelogs
+  :summary: A bundler plugin that shows changelogs of your gem dependencies that specify
+    changelog urls [not yet filtered to git version updates].
   :uri: https://github.com/jdar/bundler-changelogs
-- :name: bundler-audit
-  :uri: https://github.com/rubysec/bundler-audit
+- :name: bundler-commentate
+  :summary: Bundler plugin to add gem summaries to a Gemfile
+  :uri: https://gitlab.com/fjc/bundler-commentate
+- :name: bundler-console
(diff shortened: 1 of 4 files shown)


olleolleolle (approved):
Thanks, @ccutrer! This is a thoughtful and nice improvement.

tnir:
@ccutrer Thanks. @olleolleolle Thanks for merging. We need to copyedit the new page. I will be back here when I am available 😀

data/known_plugins.yml
```diff
@@ -1,21 +1,81 @@
...
+  :uri: http://github.com/AndyObtiva/bundler-download
 - :name: bundler-ecology
+  :summary: Bundler plugin to avoid installing unwanted gems
   :uri: https://github.com/eco-rb/bundler-ecology
 - :name: bundler-explain
+  :summary: Explains bundle update conflicts
   :uri: https://github.com/jhawthorn/bundler-explain
+- :name: bundler-fast_git
+  :summary: Make bundler git source faster
+  :uri: https://github.com/mtsmfm/bundler-fast_git
```
deivid-rodriguez, line 36: I think we may want to remove both `bundler-explain` and `bundler-fast_git` from this list since both are now included by default with Bundler.

* `bundler-explain` uses resolution errors from PubGrub, which Bundler 2.4 uses by default.
* `bundler-fast_git` uses faster git commands, which Bundler 2.4 also uses by default.

olleolleolle: Let's make an in-code data list which is about "known plugins we keep off our main list".

olleolleolle: I created https://github.com/rubygems/bundler-site/issues/1231 to have an Issue to work to.

deivid-rodriguez: Thanks, there's an `skipped_gems` list already that we may want to use.

deivid-rodriguez: Setting as resolved since this is now tracked as an issue!

PR #1232, opened 2023-10-09
Skip two Bundler plugins known to be outdated

### LINKED ISSUE #1231: known_plugins rake task: exclude `bundler-explain` and `bundler-fast_git` - both are now included by default with Bundler.
> I think we may want to remove both `bundler-explain` and `bundler-fast_git` from this list since both are now included by default with Bundler.
>
> * `bundler-explain` uses resolution errors from PubGrub, which Bundler 2.4 uses by default.
> * `bundler-fast_git` uses faster git commands, which Bundler 2.4 also uses by default.

_Originally posted by @deivid-rodriguez in https://github.com/rubygems/bundler-site/pull/1222#discussion_r1350055694_

olleolleolle:

The problem was that two plugins were seen which ought to have been skipped.

Their functionality's now built-in.

My diagnosis was that the skip list was not updated.

My fix was to extend the skip list with two more gems.

Also, I ran the Rake task, to get the result seen in data/.

I chose this fix because I got a good hint.

Fixes #1231

### DIFF
--- a/data/known_plugins.yml
+++ b/data/known_plugins.yml
@@ -28,12 +28,6 @@
 - :name: bundler-ecology
   :summary: Bundler plugin to avoid installing unwanted gems
   :uri: https://github.com/eco-rb/bundler-ecology
-- :name: bundler-explain
-  :summary: Explains bundle update conflicts
-  :uri: https://github.com/jhawthorn/bundler-explain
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Thanks @olleolleolle! ❤️

PR #1234, opened 2023-10-09
Skip bundler-shellsplit-plugin

ccutrer:

Its source does not exist, and its summary explicitly says the problem has been fixed in Bundler 2.1.0

The gem should not be listed.

Skip it.

Precedent of skipping other outdated plugins.

### DIFF
--- a/lib/tasks/regenerate_known_plugins_yml.rake
+++ b/lib/tasks/regenerate_known_plugins_yml.rake
@@ -10,11 +10,12 @@ task :regenerate_known_plugins_yml do
     extended_bundler-errors
   ]
   skipped_gems = %w[
+    bundler-explain
+    bundler-fast_git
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
Thank you!

lib/tasks/regenerate_known_plugins_yml.rake
```diff
@@ -10,11 +10,12 @@ task :regenerate_known_plugins_yml do
...
   ]
   skipped_gems = %w[
+    bundler-explain
+    bundler-fast_git
     bundler-interactive source-does-not-exist yanked-all-but-last
     bundler-next
     bundler-security
-    bundler-explain
-    bundler-fast_git
+    bundler-shellsplit-plugin
```
olleolleolle, line 18: Thanks for sorting!

PR #1239, opened 2023-10-19
Add CHANGELOG as vendored file

### LINKED ISSUE #846: 404: /changelog.html
Linked from: https://bundler.io/doc/contributing/bug_triage.html

### LINKED ISSUE #1233: Include the CHANGELOG for the release
This Issue is about making it harder to break a link in the BUG_TRIAGE section.

https://github.com/rubygems/rubygems/pull/7041#issuecomment-1752825512

Suggestion: include the rubygems/rubygems repo's Bundler CHANGELOG.md for the version we build for, and let it be called "CHANGELOG.md" in all of bunder-site.

(This is a duplicate of https://github.com/rubygems/bundler-site/issues/846.)

olleolleolle:

The problem was that a link to the CHANGELOG was 404ing. So, (in a now-closed PR) I made the link absolute, instead of relative, to work around the problem.

But, that'd take us to another website to look at Bundler's CHANGELOG. It'd be nice if Bundler's website knew about the CHANGELOG - had it locally.

My diagnosis was we didn't have all files that would be necessary to create a local CHANGELOG, we'd have to point outside this repo, to some on-GitHub data, not very nice. Not rendered in this nice way.

Include CHANGELOG as a file in the generated site.

This fixes #1233.
This fixes #846 

I chose this fix because we mentioned it in #1233.

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -7,7 +7,7 @@ config[:current_version] = config[:versions].last
 activate :syntax
 activate :i18n
 activate :search do |search|
-  search.resources = ["index.html", "guides/", "#{config[:current_version]}/", "compatibility.html", "conduct.html", "contributors.html"]
+  search.resources = ["index.html", "guides/", "#{config[:current_version]}/", "changelog.html", "compatibility.html", "conduct.html", "contributors.html"]
 
   search.index_path = "search/lunr-index.json"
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I built a review app for this and the layout does not look good. I think you need to configure it here: https://github.com/rubygems/bundler-site/blob/9cbd24f55698ac9b30cf181d68d4689a26289f4c/config.rb#L190.

Also probably make it searchable here: https://github.com/rubygems/bundler-site/blob/9cbd24f55698ac9b30cf181d68d4689a26289f4c/config.rb#L10.

olleolleolle:
Now, with the two-column layout, much easier to read:
bild

deivid-rodriguez:
Amazing @olleolleolle, thanks so much!

PR #1257, opened 2023-12-12
Back to nronn

deivid-rodriguez:

Using the fork released to rubygems.org is what @tnir did initially at https://github.com/rubygems/bundler-site/pull/841, but I reverted that at https://github.com/rubygems/bundler-site/pull/1021 because:

* I wasn't sure if it was intentionally permanent or just to unlock deployments.
* I wanted to be ronn-ng one last chance.

* ### What was your diagnosis of the problem?

My diagnosis was that ronn-ng has not really got any traction and I decided to move on to nronn in Bundler at https://github.com/rubygems/rubygems/pull/7227.

My fix is to also migrate to nronn here.

I chose this fix because it sounds like the best path going forward.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ gem "rake"
 gem "faraday-retry", "~> 2.2"
 gem "octokit", "~> 8.0"
 ## To generate ERB files from ronn files from rubygems/rubygems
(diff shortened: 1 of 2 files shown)


indirect:
Thanks for taking care of this!

PR #991, opened 2022-12-29
Add Bundler 2.4 to documentation

deivid-rodriguez:
### What was the end-user problem that led to this PR?

The problem was that I want to hold just a bit on publish the blog post, but that shouldn't block publishing documentation pages for Bundler 2.4.

### What was your diagnosis of the problem?

My diagnosis was that we should extract documentation from #978.

### What is your fix for the problem, implemented in this PR?

My fix extract documentation to this PR. Include a bare "What's new page" with just a link to the changelog for now.

### Why did you choose this fix out of the possible options?

I chose this fix because it unblocks @tnir :)

PR #1263, opened 2023-12-26
rel: add Bundler 2.5 as the latest version to docs

tnir:

Bundler 2.5.0 was successfully released on 2023-12-15, and it is still hard for end-users to find Bundler 2.1 to be fully deprecated.

- https://rubygems.org/gems/bundler/versions/2.5.0

- The latest should be Bundler 2.5.

- This publishes Bundler 2.5 as the latest version.
- This also marks Bundler 2.1 as legacy.

bundler-2 5

No other possible options.

### Refs.

- Previous update (bundler 2.4): #991

### DIFF
--- a/helpers/config_helper.rb
+++ b/helpers/config_helper.rb
@@ -14,9 +14,9 @@ module ConfigHelper
   private
 
   def status(version)
-    if version == "v2.4"
+    if version == "v2.5"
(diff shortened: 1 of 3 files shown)


deivid-rodriguez (approved):
Thank you! ❤️

PR #1279, opened 2024-01-26
rails setup: add missing newline from markdown

RaghavSood:
A missing newline causes the instructions for two different parts to merge, adding it back make things clearer.


olleolleolle (approved):
Thank you, @RaghavSood!

PR #1291, opened 2024-02-06
Document `file` option for `ruby` directive

albertchae:
Feature was implemented in https://github.com/rubygems/rubygems/pull/6876 and documented on the
[manpage](https://bundler.io/v2.5/man/gemfile.5.html#VERSION-required-)
but it seems like it makes sense to have on [this page](https://bundler.io/guides/gemfile_ruby.html) as well, as it is the top google result for "gemfile ruby version".

This copy is lifted (and lightly edited) from the manpage source https://github.com/rubygems/rubygems/blob/6bbc8bed344eeaadcbd8dfcc6e20c9e1eb734d96/bundler/lib/bundler/man/gemfile.5.ronn?plain=1#L72-L80

### DIFF
--- a/source/guides/gemfile_ruby.html.md
+++ b/source/guides/gemfile_ruby.html.md
@@ -21,6 +21,18 @@ It's also possible to restrict the patchlevel of the Ruby used by doing the foll
 ruby '1.9.3', :patchlevel => '448'
 ~~~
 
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
I think this repetition is good!

PR #1306, opened 2024-03-08
Fix typo in groups.html.md

drcapulet:


olleolleolle (approved):
Thanks, @drcapulet!

olleolleolle:
There's a way to regenerate files, it's a little Rake task, should you wish to do that, too, otherwise say "no, thanks".

drcapulet:
@olleolleolle I feel like I'm missing something - the README notes to run `bundle exec middleman build` which generates into `./build` but that's in the `.gitignore`?

olleolleolle:
@drcapulet Ah, there is a rake task `build`: https://github.com/rubygems/bundler-site/blob/master/Rakefile#L4

...which does a few things, among them, I think (I am preparing my laptop to do this, now), a set of scripts to build man files and so on, so that they have this spelling fix, too.

olleolleolle:
I am not going to hold your change up any longer, but the `rake build` task will need some looking-over.

PR #1402, opened 2024-11-25
Expand Gemfile.lock management recommendations

srizzo:

The problem was that the current Gemfile.lock recommendations for gem development were incomplete and could lead to situations where gems are published with outdated or incompatible dependencies. Users were experiencing confusion about best practices, particularly around CI processes and dependency management across different Ruby versions.

My diagnosis was that the documentation needed to address three key gaps:
1. The limitation of using a single Gemfile.lock with multiple Ruby versions
2. The timing issues with scheduled dependency checks (both Dependabot and daily CI builds)
3. The importance of dependency verification during the gem publishing process

My fix adds three important clarifications to the documentation:
1. A note about maintaining separate Gemfiles for different Ruby versions when using Dependabot
2. A recommendation to regenerate Gemfile.lock before publishing
3. More context around when and why to verify dependencies

I chose this fix because:
1. It maintains the existing recommendations while adding crucial missing context
2. It addresses real-world scenarios that developers encounter when managing gems
3. It provides actionable guidance for the gem publishing process
4. It helps prevent the release of gems with outdated or incompatible dependencies
5. The changes are concise yet comprehensive enough to guide developers toward better practices

### DIFF
--- a/source/guides/faq.html.md
+++ b/source/guides/faq.html.md
@@ -184,9 +184,7 @@ Instead of forcing every fresh checkout (and possible new
 contributor) to encounter broken builds, the Bundler team recommends
 either using a tool like [Dependabot](https://dependabot.com)
 to automatically create a PR and run the test suite any time
-your dependencies release new versions. You might need to keep 
-separate Gemfiles for different ruby versions for Dependabot dependency
-resolution to work accurately.
+your dependencies release new versions.
 If you don't want to use a dependency monitoring bot, we suggest
 creating an additional daily CI build that deletes the Gemfile.lock
 before running `bundle install`. That way you, and others monitoring


deivid-rodriguez:
Hi, thank you for trying to improve documentation.

Let's start with the first bullet point: "The limitation of using a single Gemfile.lock with multiple Ruby versions". Do you have a more elaborate explanation of this limitation? Maybe this is no longer an issue in your case with https://github.com/rubygems/rubygems/pull/8251? Before documenting workarounds, I'd like to make sure we can't provide better solutions.

srizzo:
> Hi, thank you for trying to improve documentation.
> 
> Let's start with the first bullet point: "The limitation of using a single Gemfile.lock with multiple Ruby versions". Do you have a more elaborate explanation of this limitation? Maybe this is no longer an issue in your case with [rubygems/rubygems#8251](https://github.com/rubygems/rubygems/pull/8251)? Before documenting workarounds, I'd like to make sure we can't provide better solutions.

Thank you for your quick reply. As of now, as far as I understand, different ruby versions can yield different Gemfile.lock. This might change with the PR you mentioned, but this is currently not the case.

As far as I understand, from https://github.com/dependabot/dependabot-core/issues/4987, at the moment dependabot will simply use the lower bound ruby version. 

Is that right?

srizzo:
If you can have a single Gemfile.lock for all ruby versions and Dependabot play well with that, that will be amazing, thanks for the PR.

However, at the moment, it might lead to issues with ruby versions if I understand it correctly.

I'm also happy to rephrase the PR in a way that better fits your plans, or create another one to update it once your PR is merged

deivid-rodriguez:
> I'm also happy to rephrase the PR in a way that better fits your plans, or create another one to update it once your PR is merged

I'd say let's wait for next version to be released with those improvements, then you try it and we reevaluate the situation and discuss whether anything (either docs or functionality) needs improving.

srizzo:
> Yes, that's the idea. Would you be willing to try out the PR and see if it works for your case? Also, it'd be nice to see how it works with Dependabot too.

I don't have a specific case other than a large client with dozens of (private) gems to maintain and CIs that IMHO are not optimal for the reasons I'm trying to clarify in this PR.

We did have a recent case were a Gemfile.lock had different results across ruby versions (as expected, before the PR). I can try and get more details, but I can't promise to do so in a timely manner, unfortunately.

indirect:
This note seems like a pure improvement to the docs, to me. Thanks for taking the time to write it and explain your reasoning!

srizzo:
> Cool, can you update the PR to only mention the point about testing without a lockfile before pushing a release, @srizzo?

Thank you. I've updated it.

olleolleolle:
@srizzo Fantastic, thanks for staying with it, now merged!

PR #1415, opened 2024-12-16
Fix duplicated "and"

deivid-rodriguez:

The problem was I noticed this typo while working on Bundler 2.6 updates.

My fix is to remove the duplicated word.


olleolleolle (approved):
Excellent!

PR #1421, opened 2025-01-03
Bump Ruby to 3.4 and use middleman from main branch

deivid-rodriguez:

The problem was I want to keep our stack up to date.

I contributed several PRs to middleman's main branch, so that we can keep our dependencies more up to date. The 5.x branch of middleman is abandoned anyways, and main should have all the changes that we need.


olleolleolle (approved):
Yes!

PR #1423, opened 2025-01-07
Fix CONTRIBUTING docs links

deivid-rodriguez:

The problem was broken links.

My fix is to make them not 404.


olleolleolle (approved):
Thanks! ❤️

Issue #1347, opened 2024-05-30
Url fragment is lost due to redirection


runephilosof-abtion:
On https://bundler.io/guides/using_bundler_in_applications.html#gemfilelock

https://github.com/rubygems/bundler-site/blob/4121bcabc4f88c6f413c93af14c87e939012dc32/source/localizable/guides/using_bundler_in_applications.en.html.md?plain=1#L224

The link to `/man/gemfile.5.html#PLATFORMS` will be redirected to `/v2.5/man/gemfile.5.html`, by the backend https://github.com/rubygems/bundler-site/blob/4121bcabc4f88c6f413c93af14c87e939012dc32/config.rb#L185

So the fragment `#PLATFORMS` will never have an effect.

Could it link to the current version instead?

deivid-rodriguez:
Indeed. I'm unsure how to fix this. We could make this guide use a haml template instead of markdown and dinamically link to the latest version, or hardcode the latest version in the target URL, and make sure we update it every time we release a new minor version of Bundler. None of them feel great :(

runephilosof-abtion:
I just realized that browsers of course persist the `#PLATFORMS` part, when they are redirected.
However, it seems the backend is not redirecting with a http status 302 (or 307), but instead responding with 200 OK and this body.
```
              <html>
                <head>
                  <link rel="canonical" href="/v2.6/man/gemfile.5.html" />
                  <meta http-equiv=refresh content="0; url=/v2.6/man/gemfile.5.html" />
                  <meta name="robots" content="noindex,follow" />
                  <meta http-equiv="cache-control" content="no-cache" />
                </head>
                <body>
                </body>
              </html>
```

If the backend would instead use a redirection status and location header, it should work.

runephilosof-abtion:
I can't find a way to get github pages to send a redirection response.

deivid-rodriguez:
I think middleman always uses client side redirects. This seems bad for SEO and I think it explains the issues in https://github.com/rubygems/bundler-site/issues/1333.

I think we can fix both this and https://github.com/rubygems/bundler-site/issues/1333 by not using middleman `redirect` and instead creating symlinks between `/man/` and `v2.6/man` (updated every time we bump the major version). We will serve duplicated content, but our canonical `meta` tags should keep Google happy I believe.

deivid-rodriguez:
I tried implementing this approach in #1424.

Issue #1333, opened 2024-04-18
bundler.cn ?


andyw8:
I'm seeing bundler.cn appear in Google search results, higher than bundler.io for some terms.

It's partially translated but most links still point to the normal Bundler URLs.

I assume bundler.cn isn't sanctioned, although it's been around since 2018:

https://web.archive.org/web/20181015000000*/https://www.bundler.cn/

There's probably nothing that can be done, so feel free to close, but I thought it worth a mention.

andyw8:
> I think GitHub is banned in China

[Seems it's not](https://www.reddit.com/r/programming/comments/14yhuan/comment/jrt5npg/).

deivid-rodriguez:
Alright, changes in #1424 had the desired effect and now Google ranks `bundle check` official page first in the list, and `bundler.ch` is relegated out of the first page of results 🥳.

Closing this!

PR #1424, opened 2025-01-07
Fix redirects including URL fragments

deivid-rodriguez:

The problem was that some pages include links to URLs with a fragment, and without a `v2.x/` url segment, and the fragment is actually lost when clicking the link. See for example the https://bundler.io/man/gemfile.5.html#PLATFORMS link in https://bundler.io/guides/using_bundler_in_applications.html#gemfilelock.

My diagnosis was that middleman uses client side redirects to implement redirections, so pages like https://bundler.io/man/gemfile.5.html actually serve the following content, with a 200 status:

```
curl https://bundler.io/man/gemfile.5.html 
              <html>
                <head>
                  <link rel="canonical" href="/v2.6/man/gemfile.5.html" />
                  <meta http-equiv=refresh content="0; url=/v2.6/man/gemfile.5.html" />
                  <meta name="robots" content="noindex,follow" />
                  <meta http-equiv="cache-control" content="no-cache" />
                </head>
                <body>
                </body>
              </html>
```

That means requesting a client side redirect (without the fragment).

My fix is to actually serve the same content as `v2.6/man/gemfile.5.html`, by making everything in `v2.6/man/` a symlink to `man/`. We will serve duplicated content, but our canonical meta tags should hint crawlers about what the canonical page is.

I chose this fix because it fixes https://github.com/rubygems/bundler-site/issues/1347, and it may also help with SEO issues like https://github.com/rubygems/bundler-site/issues/1333.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,6 @@ source/conduct.html.md
 source/doc
 
 # man pages built from the bundler repo
-/source/man/
 /source/v*/man/
 
 # deploy key
--- a/config.rb
+++ b/config.rb
@@ -177,14 +177,6 @@ end
   redirect "v#{version}/guides/bundler_2_upgrade.html", to: "guides/bundler_2_upgrade.html"
 end
 
-# Redirect to the latest man page when version isn't specified
-man_files = Dir.glob("./source/#{config[:current_version]}/man/*.html.erb")
-man_files.each do |file|
-  url = file.delete_prefix("./source/#{config[:current_version]}/").delete_suffix(".erb")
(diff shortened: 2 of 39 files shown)


olleolleolle:
This looks nice, and the solutions make more sense.

(I read the commits one by one, so that I could see the changelog stuff in separation.)

Let's use this!

deivid-rodriguez:
Ok to merge @olleolleolle?

deivid-rodriguez:
Cool! I'll check Google in a couple of weeks to see if something has changed regarding #1333.

PR #1425, opened 2025-01-10
Upgrade dependencies to fix most warnings

deivid-rodriguez:

The problem was that a lot of warnings are printed when building and serving the site locally.

I fixed the warnings in upstream gems, and upgraded our Gemfile.

Only one warning is missing, fixed by https://github.com/TwP/servolux/pull/22, but I can't easily point to that PR because that repository does not include a gemspec.


olleolleolle (approved):
Great!

PR #918, opened 2022-09-13
haml 6.0.0.beta.2 does not work with middleman even 5.x

tnir:
HAML 6.x (at least [haml 6.0.0.beta.2](https://rubygems.org/gems/haml/versions/6.0.0.beta.2) or haml 6.0.1) does not work with middleman even 5.x (unreleased) since middleman-syntax latest (3.3.0) does not support haml 6.x (see https://github.com/middleman/middleman-syntax/issues/80).

```
/home/runner/work/bundler-site/bundler-site/vendor/bundle/ruby/3.1.0/gems/middleman-syntax-3.3.0/lib/middleman-syntax/haml_monkey_patch.rb:4:in `<module:Haml>': Filters is not a module (TypeError)
```

### Refs.

- https://github.com/middleman/middleman-syntax/issues/80

PR #1426, opened 2025-01-15
Update to HAML 6

deivid-rodriguez:

The problem was none. I just wanted to finish the upgrade of all our stack to latest versions.

My fix is to upgrade HAML and take care of a couple of problems that showed up.

Supersedes #918.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ gem "middleman-syntax"
 
 ## Template engines
 gem "builder"
-gem "haml", "~> 5.2.2"
+gem "haml", "~> 6.3"
 gem "kramdown"
 
 # Rake tasks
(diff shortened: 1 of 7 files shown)


olleolleolle:
Code changes look like they simplify for the reader.

tnir (commented):
@deivid-rodriguez Thanks for following my progress.

While I am not able check all pages today, there is a small regression at the top page as below:

haml6

- Source https://github.com/rubygems/bundler-site/blob/cf57e550ade5164282539fb15676c9d72dd2ce98/locales/en.yml#L10

deivid-rodriguez:
Fixed, thank you!

deivid-rodriguez:
I'll merge now, we can address anything else as follow ups!

tnir:
@deivid-rodriguez Thanks for finishing up!

deivid-rodriguez:
@tnir Thanks for getting this started!

PR #1428, opened 2025-01-15
Gitignore changelog

deivid-rodriguez:

The problem was changes in this file every time a new version is released.

My diagnosis was that this file is copied from the rubygems/rubygems repository. There's no need to keep it source controlled here.

My fix is to gitignore it.

I chose this fix because it makes it easier to work with this repo.


olleolleolle (approved):
Nice, we get this information otherwise, as you stated in the PR description.

PR #1429, opened 2025-01-15
Gitignore contributors data

deivid-rodriguez:

The problem was while working on this repo, you may get random source control changes whenever the rubygems/rubygems repo gets new contributions. This is inconvenient.

My diagnosis was that there's no need to keep this file source control. We can let the repo setup generate it.

My fix is to let this file be gitignored.

I chose this fix because it makes developing in this repo nicer.


Rakefile
```diff
@@ -1,7 +1,7 @@
 Dir.glob("lib/tasks/**/*.rake").each { |r| load r }
 
 desc "Build the static site"
-task build: ["contributors:update", :repo_pages, :man] do
+task build: ["contributors", :repo_pages, :man] do
```
olleolleolle, line 4: Very minor, now possible to use a symbol here:

```suggestion
task build: [:contributors, :repo_pages, :man] do
```

olleolleolle (approved):
Great context in the PR description, thanks!

PR #1431, opened 2025-01-17
Fix transformation of some relative links

### LINKED ISSUE #844: 404: /doc/playbooks/MERGING_A_PR.md
Linked from https://bundler.io/doc/readme.html

deivid-rodriguez:

The problem was some links in the doc folder imported from rubygems/rubygems were not being properly adapted to the site structure.

My diagnosis was that the regexp to detect relative links not starting with http was incorrect, because it would fail to detect any link with target starting with "h", "t", or "p" as relative.

My fix is to properly use a negative lookahead.

This PR fixes #844. It depends on #1430 since that restore the whole page hierarchy under `/doc`.

### DIFF
--- a/lib/tasks/vendor_files.rake
+++ b/lib/tasks/vendor_files.rake
@@ -8,7 +8,7 @@ require "pathname"
 # [Heroku support](http://www.heroku.com/support)
 RELATIVE_LINK_REGEX = %r{
   \[(?<title>.*)\]            # Match title
-  \((?<link>[^(http)].*?)\)   # Match all non-http links
+  \((?<link>(?!http).*?)\)    # Match all non-http links
 }x
 
 def new_link(file, link)


lib/tasks/vendor_files.rake
```diff
@@ -30,8 +30,12 @@ end
 
 def write_file(file, to)
   content = File.read(file)
+
+  file = file.gsub("doc/bundler/", "doc/").downcase
+  to = to.gsub("doc/bundler/", "doc/").downcase
```
olleolleolle, line 35: Q: Would this be as good?

```suggestion
  file = file.sub("doc/bundler/", "doc/").downcase
  to = to.sub("doc/bundler/", "doc/").downcase
```

olleolleolle (approved):
This is nicer, having fewer absolute links in there is super.

PR #1433, opened 2025-01-20
Make sure all searchable pages have a frontmatter title

### LINKED ISSUE #1241: Search for "Pledge" does not find the Code of Conduct
Problem: I was not able to use the search box and hit the word "Pledge", which is in the Code of Conduct document.

This `config.rb` stanza

`search.resources = ["index.html", "guides/", "#{config[:current_version]}/", "changelog.html", "compatibility.html", "conduct.html", "contributors.html"]`

is perhaps not good for "vendored documents"?

deivid-rodriguez:

The problem was some content is not searchable.

My fix is to make sure it's indexed by middleman-search by making sure it has a frontmatter title.

Fixes #1241.

### DIFF
--- a/lib/tasks/vendor_files.rake
+++ b/lib/tasks/vendor_files.rake
@@ -28,7 +28,7 @@ def new_link(file, link)
   end
 end
 
-def write_file(file, to)
+def write_file(file, to, title: nil)
   content = File.read(file)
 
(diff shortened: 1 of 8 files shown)


olleolleolle (approved):
🌟 Lovely! 🌟

PR #1437, opened 2025-01-22
Point middleman back to their main branch

deivid-rodriguez:

The problem was none, rather than preferring to point to the main line rather than to a PR. 

Use the latest main branch of middleman.


olleolleolle (approved):
👏 @deivid-rodriguez Thanks!

PR #1438, opened 2025-01-22
Restore new line behavior with GFM parser

deivid-rodriguez:

The problem was after switching to GFM markdown parser in #1436, our new lines got messed up:

Captura de pantalla 2025-01-22 a las 10 28 08

My fix is to restore the previous appearance by passing the specific option to control line wrap behavior:

Captura de pantalla 2025-01-22 a las 10 28 26

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -38,7 +38,8 @@ set :markdown,
     smartypants: true,
     strikethrough: true,
     superscript: true,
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
Thanks for a super-clear PR!

PR #1440, opened 2025-01-23
Improve version selector

deivid-rodriguez:

The problem was that we browsing bundler CLI command documentation, and using the version selector, you'll be redirected to `bundle install` page if the current command is not present in the chosen version.

That is unexpected to me, since the intention when performing that action is to browse the documentation for the same command, but a different version of Bundler.

My fix is only include versions in the selector that have the current command, and to not show the selector at all if the current version is the only one defining the command.

### DIFF
--- a/source/partials/_commands_sidebar.haml
+++ b/source/partials/_commands_sidebar.haml
@@ -6,7 +6,7 @@
     %optgroup{label: status}
       - _versions.each do |version|
         - selected = version == current_visible_version
-        - value = documentation_path(current_page_without_version, version) || documentation_path('bundle-install.1', version)
+        - value = documentation_path(current_page_without_version, version)
         %option{selected: selected, value: value}
           = version
 %h4 General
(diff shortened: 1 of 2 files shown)


source/partials/_commands_sidebar.haml
```diff
@@ -1,14 +1,15 @@
...
-        - selected = version == current_visible_version
-        - value = documentation_path(current_page_without_version, version) || documentation_path('bundle-install.1', version)
-        %option{selected: selected, value: value}
-          = version
+- if versions_with_documentation_page.size > 1
+  %h4 Choose version
+  %select.version-selects.form-select.mb-3
+    - versions_with_documentation_page.reverse.group_by { status(_1) }.each do |status, _versions|
+      %optgroup{label: status}
+        - _versions.each do |version|
```
olleolleolle, line 8: Perhaps the underscored name here is a little confusing to a reader that MAY think it marks an unused variable?

deivid-rodriguez: Yes, it was already like that and I got confused about that too, indeed. Will change it 👍.

deivid-rodriguez: Done in https://github.com/rubygems/bundler-site/pull/1440/commits/c79e67101fd4226ccfa1154a594de435d20ec254!

PR #1446, opened 2025-01-28
Update default branch in deploy workflow to main

### LINKED ISSUE #1441: Rename default branch to main
We have delayed doing this in https://github.com/rubygems/rubygems due to time & technical constraints (see https://github.com/orgs/rubygems/discussions/5145), but I think doing it in this repo is straightforward, so I'd like to do it.

deivid-rodriguez:

The problem was that I renamed the default branch to main as per #1441, but the deploy workflow needs to be updated.

My fix to use the new name in the workflow.

Closes #1441.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,7 +1,7 @@
 name: Middleman deploy
 on:
   push:
-    branches: [master]
+    branches: [main]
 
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Deployment failed here because there was a branch protection in the deployment environement that did not get updated with the rename. I renamed it manually and will try again now with another PR.

PR #1448, opened 2025-01-28
known plugins: add bundler-versions_report

igneus:

There's a potentially useful plugin, but it's not yet listed on the Known Plugins page, making it virtually invisible to the end user.

The plugin was too new to be known and listed.

I added a record to the respective YAML data file.

It's the most straightforward one of all I can think of.


deivid-rodriguez (approved):
Thank you!

simi (approved):

PR #1456, opened 2025-02-11
Redirect old issues page to bug report template

deivid-rodriguez:

The problem was that issues.html page is 404.

My diagnosis was that the issues page was removed from the source repo at https://github.com/rubygems/rubygems/pull/4980.

My fix is to link to the bug report template, since it's the most similar resource that we have.

I chose this fix because it fixes the 404.


olleolleolle (approved):
Wise change!

PR #1460, opened 2025-02-13
Fix command categorization

deivid-rodriguez:

The problem was after #1459, commands are no longer properly categorized between "Primary Commands" and "Utilities".

My fix is to make sure url normalization to prepend "/" and append ".html" is applied consistently.

I also added other improvements to favor URLs without a version as much as possible since those are guaranteed to have docs for the latest version every time.


deivid-rodriguez:
Hopefully I got it right this time!

PR #1509, opened 2025-05-09
Use correct CHANGELOG file

hugopeixoto:

https://bundler.io/changelog.html was showing the rubygems changelog instead of the bundler changelog.

The bundler repository was archived and moved to the a rubygems subdirectory, so I guess that the website code was not updated to reflect the change, and continued reading from the root CHANGELOG.

Path changed to point to the right changelog.

I don't see any other options.

After this change:

image


olleolleolle (changes requested):
Good catch, and thank you for adding good context to the pull request description.

olleolleolle (approved):

PR #1521, opened 2025-06-14
Clarify gemspec arguments and behaviour.

simi:
I have found interesting confusion at [rest-client Gemfile](https://github.com/rest-client/rest-client/blob/2c72a2e77e2e87d25ff38feba0cf048d51bd5eca/Gemfile#L3-L7). It tries to load gemspec by filename using `name` argument. I was at first confused why that doesn't work, but I have realized the behaviour is different. Bundler loads all gemspecs at given path and filter them based on name entry in gemspec, not by filename.

Let's update the guides to explain this better.


olleolleolle (approved):
Thanks for making it more precise!

PR #1537, opened 2025-07-18
Make CoC clickable

deivid-rodriguez:

The problem was the email contact in CoC is not clickable.

My fix is to make it clickable, catching up with Catch up with https://github.com/rubygems/rubygems/pull/8849.


olleolleolle (approved):
Ah, excellent.

PR #1569, opened 2025-09-25
Make v2.7 the current release and rotate supported releases

ZimbiX:
I noticed the version dropdown on the website shows v2.7 as a 'Deprecated release'.

v2.7 is installed when I do `gem install bundler`, so this is most certainly wrong.

```
➜ gem install bundler
Successfully installed bundler-2.7.2
1 gem installed
```

~~I saw the last update to this version was made in e0d14d26ccc24faa1810cc968b20739becc04a6b, which also dropped the oldest version from 'Legacy release'. I've followed that here too.~~

Before:

After:

Screenshot 2025-09-29 at 1 26 20 pm

### DIFF
--- a/helpers/config_helper.rb
+++ b/helpers/config_helper.rb
@@ -16,7 +16,7 @@ module ConfigHelper
   def status(version)
     if version == "v2.7"
       "Current release"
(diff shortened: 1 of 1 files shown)


hsbt (approved):
👍

PR #1611, opened 2026-01-16
Use recommend syntax to run bundle install in Bundler v4

stefannibrasil:

When I installed Bundler v4 and ran `bundle`, I saw this:

```
$ bundle
In the next version of Bundler, running `bundle` without argument will no longer run `bundle install`.
Instead, the `help` command will be displayed.
If you'd like to keep the previous behaviour please run `bundle config set default_cli_command install --global`.
```

The warning message was straightforward.

Followed the deprecation warning and started using `bundle install` instead of just `bundle`.

As the deprecated message recommends. Even though there is the backwards compatibility option, I believe it would be nice to keep the docs up to date with how Bundler uses the command by default.


olleolleolle (approved):
👍

